Issue #17, opened 2020-05-21
Allow types that dont implement DekuRead / DekuWrite when a custom parser is provided


elast0ny:
```Rust
struct DekuTest {
    field_a: u8,
    #[deku(reader = "bytes_to_str(rest)")]
    field_b: String,
}
```
You should not need to implement anything on `String` as the custom reader handles the parsing

elast0ny:
This will be fixed when the branch is merged into master

Issue #13, opened 2020-05-14
Compile time bit_size checking


elast0ny:
https://github.com/sharksforarms/deku/blob/26eccca325c37b847b3c1ea7032045e26f8ca752/deku-derive/src/macros/deku_read.rs#L79

This seems like it could be fixed by using `const` functions. std::mem::size_of() is `const` which can all be evaluated at compile time. e.g. :
```Rust
    const fn bit_size() -> usize {
        std::mem::size_of::<$typ>() * 8
    }
```

sharksforarms:
I wonder if the assert could be done at compile time? For example, this would fail to compile

```rust
struct DekuTest {
    #[deku(bits = "9")]
    field_a: u8,
}
```

elast0ny:
This should be possible with :
https://docs.rs/static_assertions/1.1.0/static_assertions/macro.const_assert.html

elast0ny:
oh crap : https://github.com/rust-lang/rust/issues/57563

sharksforarms:
No longer required: https://github.com/sharksforarms/deku/pull/19

PR #19, opened 2020-05-21
Removing bit size trait

elast0ny:


elast0ny:
This PR makes issue #13 obsolete as length checking is now only done at runtime

elast0ny:
The generated macro code now only calls either `read` or `write` which fixes issue #17 as no other trait implementations must be defined on custom types with custom parsers.

PR #61, opened 2020-07-19
Attempt to add context support

constfold:
### Motivation

This PR added a generic parameter called `Ctx` for `DekuRead` and `DekuWrite` to support context transmitting. So we can replace fixed arguments `input_is_le`, `bit_size`, and `count` (only used by `Vec`) with more flexible ones. Currently we need to pass a `None` even those arguments won't be used at all.

And it will make providing an attribute of context possible:

```rust
#[deku(ctx = "a: u8")]
struct Foo{}
struct Bar {
    b: u8,
    #[deku(ctx = "a")]
    foo: Foo,
    #[deku(ctx = "a, foo")]
    foos: Vec<Xx>
}
```

### Detail

Both of `DekuRead` and `DekuWrite` have an generic parameter to allow users implement context (`Ctx`) sensitive structure:

```rust
pub trait DekuRead<Ctx> {
    fn read(input: _, ctx: Ctx) -> Self;
}
pub trait DekuWrite<Ctx> {
    fn write(&self, ctx: Ctx) -> Vec<u8>;
}
```

All primary types need two "context" (`endian` and `bit_size`):

```rust
impl DekuRead<(Endian, usize)> for u8 {
    fn read(input: _, (endian, bit_size): (Endian, usize)) -> u8{}
}
impl DekuRead<Endian> for u8 {
    fn read(input: _, endian: Endian) -> u8 { u8::read(endian, size_of<u8>()) }
}
// usage:
let _ = u8::read(Endian::Native);
let _ = u8::read(Endian::Bit, 12);
```

A `Vec` need a `count` (and some addition context for inner type):

```rust
impl<T, InnerCtx> DekuRead<(Count, InnerCtx)> for Vec<T> {
    fn read(input: _, (count, inner_ctx): (Count, InnerCtx)) -> Vec<T> {
        for _i in 0..count {
            let (new_rest, val) = <T>::read(rest, inner_ctx);
            res.push(val);
            rest = new_rest;
        }
    }
}
impl<T> DekuRead<Count> for Vec<T> {}
// usage:
let _ = Vec::<u8>::read((2, (Endian::Native, 7));
let _ = Vec::<u8>::read((2, Endian::Native));
```

#### Derive Macro

A `ctx` attribute will automatically add context for a structure: 

```rust
#[derive(DekuRead)]
#[deku(ctx = "a: u8, x: String, y: bool")]
struct Foo {
    #[deku(cond="a == 0", endian="big")]
    b: Option<u8>;
    #[deku(ctx="y", count=a)]
    c: Vec<Bar>;
}
// expanded to
impl DekuRead<(u8, String, bool)> for Foo {
    fn read(input: _, (a, x, y): (u8, String, bool)) -> Self {
        let b = if a == 0 {
            Some(u8::read(input, Endian::Big))
        } else { 
            None
        };
        let c = Vec<Bar>::read(input, (a, y));
        Foo{ b, c }
    }
}
// usage
let _ = Foo::read((1, "".to_owned(), false));
```

Why not use associate type like `DekuRead::Ctx`? Because a generic trait allows us to write specific implementation (like default value).

### Drawback

- The implementation of `Vec` is confusing, what `Vec::read((1, (2, 3, 4)))` (or ``Vec::read((VecCount(1), (2, 3, 4)))`) means?
- More complexity
- Difficult to write derive macro(e.g. Not sure how would we parse "a: u8" to `(a): (u8)`)

### Todo

A lot of work to do

- [x] Modify trait declaration
- [x] Modify derive macro
- [x] Add `ctx` attribute
- [x] Add useful types(like `Endian`)
- [ ] Rewrite document
- [ ] Add tests


src/lib.rs
```diff
@@ -617,51 +600,47 @@ mod tests {
...
-
-        #[should_panic(expected="Parse(\"not enough data: expected 32 got 0\")")]
-        case::not_enough_data([].as_ref(), IS_LE, Some(32), None, 0xFF, bits![Msb0, u8;]),
-        #[should_panic(expected="Parse(\"not enough data: expected 32 got 16\")")]
-        case::not_enough_data([0xAA, 0xBB].as_ref(), IS_LE, Some(32), None, 0xFF, bits![Msb0, u8;]),
-        #[should_panic(expected="Parse(\"too much data: container of 32 cannot hold 64\")")]
-        case::too_much_data([0xAA, 0xBB, 0xCC, 0xDD, 0xAA, 0xBB, 0xCC, 0xDD].as_ref(), IS_LE, Some(64), None, 0xFF, bits![Msb0, u8;]),
-        #[should_panic(expected="Dev error: `count` should always be None")]
-        case::dev_err_count_some([].as_ref(), IS_LE, Some(64), Some(1), 0xFF, bits![Msb0, u8;]),
+    #[rstest(input, input_is_le, bit_size, expected, expected_rest,
```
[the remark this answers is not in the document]
constfold: Ah, it's an accident. I'm using `rustfmt 1.4.18-nightly (c1e9b7b8 2020-06-13)`. Seem like the nightly version will format codes in attribute now. I would like to reset it back manually if you mind the new format.
BTW, I mentioned two formats changed:
1. A new space after a comma. I think it's fine as it was suggusted in [Rust Style Guide](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md#attributes)
2. The indentation disappeared. It makes codes less readable. Maybe it's a bug of `rustfmt`?

constfold:
@sharksforarms I partly finished this PR, but there still is so many things need to rewrite (tests, examples, the usages of attributes). So what do you think?
Note for things need to talk:
1. struct/enum level endian. It's a little diffcult. Because not all fields need an endian(maybe). What should we do if a custom type has no endian as its context?
2. endian/bit_size attribute. they are all somewhat special. (And will be passed by order(endian -> bit_size -> count))
3. New types like `VecCount`, `Endian`, `BitSize`. It makes error message and function definition more clear and readable. Using `usize`/`bool` anywhere is a little ugly i think.

src/lib.rs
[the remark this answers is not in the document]
sharksforarms: Makes sense! No worries :)

sharksforarms:
> @sharksforarms I partly finished this PR, but there still is so many things need to rewrite (tests, examples, the usages of attributes). So what do you think?

Looks great so far! Some more polishing to be done but so far so good! I agree with the changes proposed.

> 1. struct/enum level endian. It's a little diffcult. Because not all fields need an endian(maybe). What should we do if a custom type has no endian as its context?

I think it's fair to assume that the endian-ness specified in that way is invalid, for example:

Invalid:
```rust
struct CustomType {
   field_a: u16
}

struct MyStruct {
   #[deku(endian = "big")]  
   field_a: u16,

   #[deku(endian = "big")] // <-- invalid
   field_b: CustomType,
}
```

Valid:
```rust
#[deku(endian = "big")] // valid
struct CustomType {
   field_a: u16
}

struct MyStruct {
   #[deku(endian = "big")]  
   field_a: u16,

   field_b: CustomType, // valid
}
```

> 2. endian/bit_size attribute. they are all somewhat special. (And will be passed by order(endian -> bit_size -> count))

What do you mean by this? Are you referring to this? https://github.com/sharksforarms/deku/pull/61/files#diff-0b5720103b09979bf3553a247e0c0917R276

Correct me if I'm wrong, depending on which attributes are specified, it'll call the correct impl
impl <(bool)>
impl <(bool, usize)>
impl <(bool, usize, usize)>

(This will be more clear with #3)

> 3. New types like `VecCount`, `Endian`, `BitSize`. It makes error message and function definition more clear and readable. Using `usize`/`bool` anywhere is a little ugly i think.

I agree with this. Easier for documentation and readability.

constfold:
Oh I deleted those useful sentences by mistake. The thing I want to say in `#2` is: _The conflict between `endian`/`bit_size`/`count` and `ctx`._ For example:
```rust
// We need to sort ctx by order
#[deku(ctx = "endian: Endian, bit_size: BitSize, count: Count")]
struct Subtype {}
struct Container {
    #[deku(bits = 12, count = 1, endian = "big")]
    s: Subtype
}
```
When we use them with `ctx`:
```rust
// #[deku(ctx = "endian: Endian, bit_size: BitSize, count: Count")] --> Wrong!
#[deku(ctx = "count: Count, endian: Endian, bit_size: BitSize")]
struct Subtype {}
struct Container {
    endian: Endian,
    // We can't use `endian = "endian"` for now.
    #[deku(bits = 12, ctx = "endian, BitSize(1)")]
    s: Subtype
}
```
This behaviour is a little bit werid(not sure. maybe a good document will resolve it).

sharksforarms:
> Oh I deleted those useful sentences by mistake. The thing I want to say in `#2` is: _The conflict between `endian`/`bit_size`/`count` and `ctx`._ For example:
> 
> ```rust
> // We need to sort ctx by order
> #[deku(ctx = "endian: Endian, bit_size: BitSize, count: Count")]
> struct Subtype {}
> struct Container {
>     #[deku(bits = 12, count = 1, endian = "big")]
>     s: Subtype
> }
> ```

Right, I see what you mean... Does this make sense to pass through? I would be ok with a compile error if `bits`, `count`, `endian`, etc. was used on `SubType` (a type not native to deku) (can this easily be done?), if these are wanted, it could be written as:

```rust
#[deku(ctx = "endian: Endian, bit_size: BitSize, count: Count")]
struct Subtype {}
 struct Container {
     #[deku(ctx = "Endian::Native, BitSize(12), Count(1)")]
     s: Subtype
}
```

Does that make sense? Might be easier for implementation and removes the conflict.

sharksforarms:
i.r. reader/writer, we can enforce only one is used, `ctx` or `reader/writer` until a better solution is thought of

deku-derive/src/lib.rs
```diff
@@ -24,77 +26,310 @@ impl Default for EndianNess {
...
-        // Validate that `id_type` is set with a size
-        if (self.id_bits.is_some() || self.id_bytes.is_some()) && self.id_type.is_none() {
-            panic!("`id_type` must be specified with `id_bits` or `id_bytes`");
-        }
+impl DekuData {
+    /// Map `DekuReceiver` to `DekuData`. It will check if attributes valid. Return a compile error
+    /// if failed.
+    fn from_receiver(receiver: DekuReceiver) -> Result<Self, TokenStream> {
+        // Validate
+        DekuData::validate(&receiver)
```
sharksforarms, line 54: this is great!

deku-derive/src/lib.rs
```diff
@@ -24,77 +26,309 @@ impl Default for EndianNess {
...
+                } else if receiver.id_bits.is_some() {
+                    Err((receiver.id_bits.span(), "`id_bits` only supported on enum"))
+                } else {
+                    Ok(())
+                }
+            }
+            ast::Data::Enum(_) => {
+                // Validate either `id_bits` or `id_bytes` is specified
+                if (receiver.id_bits.is_some() || receiver.id_bytes.is_some())
+                    && receiver.id_type.is_none()
```
sharksforarms, line 113: I think we could validate `id_type` is set here for enums, and unwrap the option in deku-read/write. That way we're validating in one location

constfold:
Ah! The next step will be writing examples, documents and tests. And I will try to replace current `ctx` with `Count`, `Endian` etc. simultaneously.
Anyway why is this PR so big?

deku-derive/src/lib.rs
constfold: Yes, but I think put all validating code in one place is more readable. And we can call `unwrap` in `emit_enum`.

sharksforarms:
> Ah! The next step will be writing examples, documents and tests. And I will try to replace current `ctx` with `Count`, `Endian` etc. simultaneously.
> Anyway why is this PR so big?

I like the improvements to the overall structure, having compile errors vs panics is awesome.

I can help with the examples, documentation and testing if you'd like :)

constfold:
> 
> 
> > Ah! The next step will be writing examples, documents and tests. And I will try to replace current `ctx` with `Count`, `Endian` etc. simultaneously.
> > Anyway why is this PR so big?
> 
> I can help with the examples, documentation and testing if you'd like :)
Thanks! That's great.

deku-derive/src/lib.rs
constfold: Ah I understand what you mean. I will add this validation.

deku-derive/src/lib.rs
```diff
@@ -24,77 +26,309 @@ impl Default for EndianNess {
...
+    ident: syn::Ident,
+    generics: syn::Generics,
+    data: ast::Data<DekuVariantReceiver, DekuFieldReceiver>,
+
+    /// Endian default for the fields
+    #[darling(default)]
+    endian: Option<EndianNess>,
+
+    /// struct/enum level ctx like "a: u8, b: u8"
+    /// The type of it should be `syn::punctuated::Punctuated<syn::FnArg, syn::token::Comma>`. `darling`
```
sharksforarms, line 314: https://github.com/TedDriggs/darling/pull/98

deku-derive/src/lib.rs
```diff
@@ -148,6 +382,10 @@ struct DekuFieldReceiver {
     #[darling(default, map = "option_as_tokenstream")]
     map: Option<TokenStream>,
 
+    /// context like `"a, c + 1"`. We will parse it to `Punctuated<Expr, Comma>` latter.
```
sharksforarms, line 385: https://github.com/TedDriggs/darling/pull/98

deku-derive/src/macros/deku_read.rs
```diff
@@ -262,38 +279,71 @@ fn emit_field_read(
...
+
+        // Count is special, we need to generate `(count, (other, ..))` for it.
+        if let Some(field_count) = &f.count {
+            // The count has same problem, when it isn't a copy type, the field will be moved.
+            // e.g. struct FooBar {
+            //   a: Baz // a type implement `Into<usize>` but not `Copy`.
+            //   #[deku(count = "a") <-- Oops, use of moved value: `a`
+            //   b: Vec<_>
+            // }
+            quote! {DekuRead::read(rest, (usize::try_from(#field_count << 0usize)?, (#read_args)))}
```
sharksforarms, line 315: Bit confused here, why is `#field_count << 0usize` needed?

deku-derive/src/macros/mod.rs
```diff
@@ -65,3 +59,95 @@ fn gen_struct_init<I: ToTokens>(is_named: bool, field_idents: Vec<I>) -> TokenSt
...
+        }
+    }
+}
+
+fn gen_hidden_field_ident(ident: TokenStream) -> TokenStream {
+    // We can't concat to token, so I use string.
+    // See https://github.com/rust-lang/rust/issues/29599
+    let span = ident.span();
+    let s = ident.to_string();
+    let mut name = "__".to_owned();
```
sharksforarms, line 91: we could rename `hidden` to `internal` and change `__` to `__deku_` to imply internal variable

deku-derive/src/macros/deku_read.rs
```diff
@@ -102,9 +106,16 @@ fn emit_enum(input: &DekuReceiver) -> Result<TokenStream, darling::Error> {
     let ident = &input.ident;
     let ident = quote! { #ident #ty };
 
+    // TODO: replace `expect` with an error.
     let id_type = input.id_type.as_ref().expect("expected `id_type` on enum");
-    let id_is_le_bytes = input.endian == EndianNess::Little;
-    let id_bit_size = super::option_as_literal_token(input.id_bits);
+
+    let id_is_le_bytes = input.endian.unwrap_or_default() == EndianNess::Little;
```
sharksforarms, line 112: To stay consistent with others:
```rust
    let id_is_le = input
        .endian
        .or(input.endian)
        .map(|endian| endian == EndianNess::Little)
```

deku-derive/src/macros/deku_write.rs
```diff
@@ -282,35 +328,50 @@ fn emit_field_update(
...
+        .or(input.endian)
+        .map(|endian| endian == EndianNess::Little);
     let field_writer = &f.writer;
     let field_ident = f.get_ident(i, object_prefix.is_none());
 
     let field_write_func = if field_writer.is_some() {
         quote! { #field_writer }
     } else {
-        quote! { #object_prefix #field_ident.write(output_is_le, field_bits) }
+        let mut write_args = Vec::with_capacity(2);
```
sharksforarms, line 351: `(3)` ?

sharksforarms:
Do you know why it's saying `warning: unused variable: a` ?

example.rs
```rust
use deku::prelude::*;

#[deku(ctx = "a: &u8")]
struct Subtype {
    #[deku(skip, default = "a + 1")]
    data: u8,
}

struct DekuTest {
    field_a: u8,
    #[deku(ctx = "field_a")]
    sub: Subtype,
}

fn main() {
}
```

```rust
cargo run --example example
   Compiling deku v0.6.1 (/home/sharks/source/deku)
warning: unused variable: `a`
 --> examples/example.rs:4:14
  |
4 | #[deku(ctx = "a: &u8")]
  |              ^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_a`
  |
  = note: `#[warn(unused_variables)]` on by default

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.40s
     Running `target/debug/examples/example`
```

tests/test_alloc.rs
```diff
@@ -5,15 +5,17 @@ use deku::prelude::*;
...
     pub field_a: u8,
 }
 
 #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
-#[deku(id_type = "u8")]
+#[deku(id_type = "u8", ctx = "is_le_bytes: bool")]
 enum NestedEnum {
     #[deku(id = "0x01")]
-    VarA(u8),
+    VarA(#[deku(ctx = "is_le_bytes")] u8),
```
sharksforarms, line 18: are these changes intended?

tests/test_macro.rs
```diff
@@ -104,35 +104,6 @@ mod tests {
             }
         }
 
-        #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
```
sharksforarms, line 107: Can this test be re-added?

deku-derive/src/macros/deku_write.rs
```diff
@@ -130,7 +130,7 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
     let ident = quote! { #ident #ty };
 
     let id_type = input.id_type.as_ref().expect("expected `id_type` on enum");
-    let id_is_le_bytes = input.endian == EndianNess::Little;
+    let id_is_le_bytes = input.endian.unwrap_or_default() == EndianNess::Little;
```
sharksforarms, line 133: To stay consistent with others:
```rust
    let id_is_le = input
        .endian
        .or(input.endian)
        .map(|endian| endian == EndianNess::Little)
```

deku-derive/src/lib.rs
sharksforarms: Yes exactly, the validation can be done here and then unwrap later 👍

deku-derive/src/lib.rs
constfold: Great

deku-derive/src/macros/deku_read.rs
constfold: It's for type converting. Since `TryFrom` doesn't implement for &T(and `#field_cound` might be `123_i32` or `&i32`), I need to find a way to convert it to usize(`&i32 << usize == i32`). But I find here is a better way: using `usize::try_from(*(#field_cound.borrow()))`, I will fix it.

deku-derive/src/macros/mod.rs
constfold: Ah I didn't know it before. Thanks

deku-derive/src/macros/deku_read.rs
constfold: `unwrap_or_default()` and `or()` is different. the first one means "It could be None, and we will give it a default value". The second one means "We try it first, if None use other". And 
```rust
let id_is_le_bytes: bool = input.endian.unwrap_or_default() == EndianNess::Little;
let id_is_le: Option<bool> = input
        .endian
        .or(input.endian)
        .map(|endian| endian == EndianNess::Little)
```

deku-derive/src/macros/deku_write.rs
constfold: I will fix it

tests/test_alloc.rs
constfold: Yes, they are. As I mentioned in [this comment](https://github.com/sharksforarms/deku/pull/61#issuecomment-661894039), a top-level endian will imply that all fields require a `endian ctx`. So I added `ctx = "is_le_bytes: bool"` for its subtype. But because the field level `endian` (for now) only accept "big" or "little"(we want to use variant here), I added `#[deku(ctx = "is_le_bytes")] u8`.

tests/test_macro.rs
constfold: I will give it a try

src/ctx.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// Is it a little endian.
+    pub fn is_le(&self) -> bool {
+        match self {
+            Endian::Little => true,
+            _ => false,
+        }
+    }
+
+    /// Is it a big endian.
+    pub fn is_ge(&self) -> bool {
```
sharksforarms, line 43: `is_be`

deku-derive/src/macros/deku_read.rs
```diff
@@ -262,38 +279,76 @@ fn emit_field_read(
...
+            // The count has same problem, when it isn't a copy type, the field will be moved.
+            // e.g. struct FooBar {
+            //   a: Baz // a type implement `Into<usize>` but not `Copy`.
+            //   #[deku(count = "a") <-- Oops, use of moved value: `a`
+            //   b: Vec<_>
+            // }
+            quote! {
+                {
+                    use core::borrow::Borrow;
+                    DekuRead::read(rest, (usize::try_from(*((#field_count).borrow()))?, (#read_args)))
```
sharksforarms, line 318: Should be `Count`?

src/ctx.rs
constfold: Ah.... I am such an idiot.

constfold:
Congratulations. Finally The part of writng codes is almost finished.

sharksforarms:
> Congratulations. Finally The part of writng codes is almost finished.

Very exiting :) Thanks for your time and effort!!

tests/test_alloc.rs
```diff
@@ -1,19 +1,22 @@
 use alloc_counter::AllocCounterSystem;
+use deku::ctx::Endian;
```
sharksforarms, line 2: Do we want to add this to the prelude? (I'm debating, I'm leaning more towards no as users typically won't be needing it)

deku-derive/src/lib.rs
```diff
@@ -1,45 +1,296 @@
...
+use darling::{ast, FromDeriveInput, FromField, FromVariant};
 use proc_macro2::TokenStream;
 use quote::quote;
 mod macros;
 use crate::macros::{deku_read::emit_deku_read, deku_write::emit_deku_write};
+use syn::punctuated::Punctuated;
+use syn::spanned::Spanned;
+
+/// # Note
+/// We use this instead of `DekuReceiver::init` because handle everything in one struct is hard to use,
```
sharksforarms, line 10: Can remove this comment now that `DekuReceiver::init` is gone

deku-derive/src/macros/mod.rs
```diff
@@ -65,3 +59,147 @@ fn gen_struct_init<I: ToTokens>(is_named: bool, field_idents: Vec<I>) -> TokenSt
...
+
+            Ok((ctx_types, ctx_arg))
+        }
+    } else {
+        Ok((quote! {()}, quote! {_: ()}))
+    }
+}
+
+fn gen_id_args(endian: Option<&syn::LitStr>, bits: Option<usize>) -> TokenStream {
+    let endian = endian.map(|s| {
```
sharksforarms, line 156: Could we do this in `DekuData::from_receiver` as part of the validation? That way we have a valid object to work with

deku-derive/src/macros/mod.rs
```diff
@@ -65,3 +59,147 @@ fn gen_struct_init<I: ToTokens>(is_named: bool, field_idents: Vec<I>) -> TokenSt
...
+        quote! {<deku::ctx::Endian as core::str::FromStr>::from_str(#s)}
+    });
+
+    let bits = bits.map(|n| quote! {deku::ctx::BitSize(#n)});
```
sharksforarms, line 160: Possibly the same as above with bits?

deku-derive/src/macros/mod.rs
```diff
@@ -65,3 +59,147 @@ fn gen_struct_init<I: ToTokens>(is_named: bool, field_idents: Vec<I>) -> TokenSt
...
+fn gen_field_args(
+    endian: Option<&syn::LitStr>,
+    bits: Option<usize>,
+    ctx: Option<&Punctuated<syn::Expr, syn::token::Comma>>,
+) -> syn::Result<TokenStream> {
+    let mut args = Vec::with_capacity(3);
+
+    if let Some(field_endian) = endian {
+        args.push(match field_endian.value().as_str() {
+            "little" => quote! {deku::ctx::Endian::Little},
```
sharksforarms, line 186: `from_str` + map_err here?

tests/test_alloc.rs
constfold: I agree with you. IMO, most of users write codes like:
```rust
struct DataType {
    // ...
}
let data = DataType::from_bytes(input);
data.to_bytes();
```
In most case they don't need to care about context.

deku-derive/src/lib.rs
constfold: Yes. Actually it should be in a commit message, instead comment.

deku-derive/src/macros/mod.rs
constfold: At first I think the same way with you. But when I try to do it, I found I must add `deku` as a dependency.

deku-derive/src/lib.rs
```diff
@@ -1,45 +1,296 @@
...
-        #[cfg(target_endian = "little")]
-        let ret = EndianNess::Little;
+#[derive(Debug)]
+struct FieldData {
+    ident: Option<syn::Ident>,
+    ty: syn::Type,
 
-        #[cfg(target_endian = "big")]
-        let rets = EndianNess::Big;
+    /// Endianness for the field, `little` or `big`
```
sharksforarms, line 142: can remove `, little or big`

deku-derive/src/macros/mod.rs
constfold: Yeah, I will add it.

deku-derive/src/macros/mod.rs
constfold: Same as above. `#[deku(id_type = "xx")]` is allowed.

deku-derive/src/macros/deku_read.rs
```diff
@@ -38,9 +37,13 @@ fn emit_struct(input: &DekuReceiver) -> Result<TokenStream, darling::Error> {
 
     let (field_idents, field_reads) = emit_field_reads(input, &fields)?;
 
-    let initialize_struct = super::gen_struct_init(is_named_struct, field_idents);
+    let hidden_fields = gen_internal_field_idents(is_named_struct, field_idents);
```
sharksforarms, line 40: `internal_fields`

deku-derive/src/macros/deku_read.rs
```diff
@@ -134,8 +142,9 @@ fn emit_enum(input: &DekuReceiver) -> Result<TokenStream, darling::Error> {
         } else {
             let (field_idents, field_reads) = emit_field_reads(input, &variant.fields.as_ref())?;
 
+            let hidden_fields = gen_internal_field_idents(variant_is_named, field_idents);
```
sharksforarms, line 145: `internal_fields`

deku-derive/src/macros/mod.rs
sharksforarms: Ah! Makes sense.

deku-derive/src/macros/deku_read.rs
```diff
@@ -242,58 +259,78 @@ fn emit_field_reads(
...
+            quote! {DekuRead::read(rest, (#read_args))}
+        }
     };
 
+    // We must pass a ref to context so that it won't be moved when use.
+    // e.g.
+    // let a = read(rest);
+    // let b = read(rest, a); <-- Oops! a have been moved, then we can't use it for constructing.
+    // let c = read(rest, &mut b); <-- `b` will be changed.
+    // So I add a `__`(double underscore) for it. Hopes none writes `let d = read(rest, __b);`
```
sharksforarms, line 313: the double underscore section of the comment can be removed

deku-derive/src/macros/deku_write.rs
```diff
@@ -1,54 +1,93 @@
...
-fn emit_struct(input: &DekuReceiver) -> Result<TokenStream, darling::Error> {
+fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
     let mut tokens = TokenStream::new();
 
     let (imp, ty, wher) = input.generics.split_for_impl();
 
     let ident = &input.ident;
     let ident = quote! { #ident #ty };
 
+    // TODO: Replace `expect` with `Err`
```
sharksforarms, line 22: Since we know it's a struct, this is ok?

deku-derive/src/macros/deku_write.rs
```diff
@@ -1,54 +1,93 @@
...
-    tokens.extend(quote! {
-        impl #imp core::convert::TryFrom<#ident> for BitVec<Msb0, u8> #wher {
-            type Error = DekuError;
+    /*
+    NOTE:
+    Because the requirement by `ctx`, we need to deconstruct first.
+    e.g.: match *self { Self{ ref field_0, ref field_1 } => { /* do something */} }
+     */
+    // We checked in `emit_deku_write`.
+    let r#struct = input.data.as_ref().take_struct().unwrap();
```
sharksforarms, line 38: can `fields` not be re-used?

deku-derive/src/macros/deku_write.rs
```diff
@@ -1,54 +1,93 @@
...
-    let field_writes = emit_field_writes(input, &fields, Some(quote! { self. }))?;
+    let field_writes = emit_field_writes(input, &fields, None)?;
     let field_updates = emit_field_updates(&fields, Some(quote! { self. }))?;
 
-    tokens.extend(quote! {
-        impl #imp core::convert::TryFrom<#ident> for BitVec<Msb0, u8> #wher {
-            type Error = DekuError;
+    /*
+    NOTE:
+    Because the requirement by `ctx`, we need to deconstruct first.
```
sharksforarms, line 34: can you clarify the comment? why do we need to deconstruct? scoping/naming issue?

deku-derive/src/macros/deku_write.rs
```diff
@@ -1,54 +1,93 @@
...
+    let r#struct = input.data.as_ref().take_struct().unwrap();
+    let named = r#struct.style.is_struct();
+
+    let field_idents = r#struct
+        .iter()
+        .enumerate()
+        .map(|(i, f)| f.get_ident(i, true))
+        .collect::<Vec<_>>();
+
+    let destruction = gen_struct_destruction(named, &input.ident, &field_idents);
```
sharksforarms, line 47: Rename "destruction" to "destructured"

deku-derive/src/macros/deku_write.rs
```diff
@@ -1,54 +1,93 @@
...
+
+    // A type is container only if it's not required any context
```
sharksforarms, line 49: `// Only implement `DekuContainerWrite` for types don't need any context.`

deku-derive/src/macros/deku_write.rs
```diff
@@ -167,22 +198,48 @@ fn emit_enum(input: &DekuReceiver) -> Result<TokenStream, darling::Error> {
         });
     }
 
-    tokens.extend(quote! {
-        impl #imp core::convert::TryFrom<#ident> for BitVec<Msb0, u8> #wher {
-            type Error = DekuError;
+    // A type is container only if it's not required any context
+    if input.ctx.is_none() {
```
sharksforarms, line 202: ```// Only implement `DekuContainerWrite` for types don't need any context.```

src/ctx.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub fn is_le(&self) -> bool {
+        match self {
+            Endian::Little => true,
+            _ => false,
+        }
+    }
+
+    /// Is it a big endian.
+    pub fn is_be(&self) -> bool {
+        match self {
```
sharksforarms, line 44: `self == Endian::Big`

src/ctx.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        #[cfg(target_endian = "big")]
+        let endian = Endian::Big;
+
+        endian
+    }
+
+    /// Is it a little endian.
+    pub fn is_le(&self) -> bool {
+        match self {
```
sharksforarms, line 36: `self == Endian::Little`

src/ctx.rs
```diff
@@ -0,0 +1,172 @@
...
+///
+/// [`from_str`]: enum.Endian.html#method.from_str
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ParseEndianError {}
+
+impl Endian {
+    /// As same as [`Endian::default`], but const.
+    ///
+    /// [`Endian::default`]: #method.default
+    pub const fn new() -> Self {
```
sharksforarms, line 24: maybe
```rust

if cfg!(target_endian = "little") {
    Endian::Little
} else {
    Endian::Big
}
```

src/ctx.rs
```diff
@@ -0,0 +1,172 @@
+//! This module provided types for context representation could be used in context-sensitive parsing.
+//! See [ctx attribute](attributes/index.html#ctx) for more information.
+
+use core::ops::{Deref, DerefMut};
+use core::str::FromStr;
+
+/// An endian.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum Endian {
```
sharksforarms, line 9: Note: At some point we might add `Network` (which == big)
`deku(endian = "network")` for clarity

sharksforarms: Or, the `from_str` "network" -> `Endian::Big` might be better

src/ctx.rs
```diff
@@ -0,0 +1,172 @@
...
+
+impl FromStr for Endian {
+    type Err = ParseEndianError;
+
+    /// Parse a `Endian` from a string.
+    /// # Examples
+    /// ```rust
+    /// use std::str::FromStr;
+    /// use deku::ctx::Endian;
+    /// assert_eq!(FromStr::from_str("little"), Ok(Endian::Little));
```
sharksforarms, line 66: Any reason to write like this instead of:
```rust
    /// assert_eq!(Endian::from_str("little"), Ok(Endian::Little));
    /// assert_eq!(Endian::from_str("big"), Ok(Endian::Big));
    /// assert!(Endian::from_str("not an endian").is_err());
```

src/ctx.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// # Examples
+    /// ```rust
+    /// # use std::mem::size_of;
+    /// # use deku::ctx::BitSize;
+    ///
+    /// assert_eq!(BitSize::with_byte_size(1), BitSize(8));
+    /// ```
+    ///
+    /// # Panic
+    /// Panic if `byte_size * 8` is greater than `usize::MAX`.
```
sharksforarms, line 124: won't panic in --release mode (debug mode does panic)

sharksforarms: we can do a `checked_mul` and expect on the option

src/lib.rs
```diff
@@ -236,29 +231,46 @@ pub trait DekuUpdate {
     fn update(&mut self) -> Result<(), DekuError>;
 }
 
+/*
```
sharksforarms, line 234: Comment can now be removed

src/lib.rs
```diff
@@ -456,72 +531,69 @@ ImplDekuTraits!(isize);
 ImplDekuTraits!(f32);
 ImplDekuTraits!(f64);
 
+/*
+FIXME: Currently those implementation will expose it's implementation detail.
```
sharksforarms, line 535: Confused what you mean by this comment

deku-derive/src/macros/deku_write.rs
constfold: it's a naming issue. for example:
```rust
#[derive(DekuWrite)]
struct A {
    a: u8,
    #[deku(ctx = "a")]
    b: Subtype,
    #[deku(writer = "write((a, b))")]
    c: u8
}
// expand to
fn write(...) -> ... {
    self.b.write(a); // <-- `a` not found, should be &self.a
    write((a, b)); // <-- `a, b` not found, should be &self.a, &self.b
}
```

src/ctx.rs
constfold: Just one unimportant reason: `FromStr::from_str` is more right since in Rust you can write:
```rust
struct A();
impl A {
    fn from_str() {}
}
impl FromStr for A {
    fn from_str(_: &str) -> ... {}
}
let a = A::from_str("a string"); // compile error. `A::from_str` exactly ref to `fn from_str() {}`.
let a: A = FromStr::from_str("a string"); // good
```

src/lib.rs
constfold: I will remove it. It means a user can know it will read a u32 through the bound `u32: DekuRead<Ctx>`. It's find because everyone knows Ipv4 is u32 and Ipv6 is u128

src/ctx.rs
constfold: We can mark Endian `#[non_exhaustive]` then add a new variant called `Network` to `enum Endian` and treat it as `Big`.(or other way) I dislike the idea of "network" -> `Endian::Big`.

src/lib.rs
constfold: It could be done in https://github.com/sharksforarms/deku/pull/61/files/5e7deecaeda49f31229b2d691aa0367791a3485d#diff-3227a35819c9da6bf1bf643fad7e99a8R186

src/ctx.rs
constfold: It could be done in https://github.com/sharksforarms/deku/pull/61/files/5e7deecaeda49f31229b2d691aa0367791a3485d#diff-3227a35819c9da6bf1bf643fad7e99a8R186

src/ctx.rs
constfold: Ah can't do it `new` is const fn so `if` is nt allowed

constfold:
Ahhhhh so many things need to test. I even don't know what thing is acceptable what is not.

deku-derive/src/macros/deku_write.rs
sharksforarms: Ah! Gotcha.

src/ctx.rs
sharksforarms: Ah interesting! I'm surprised there's no clippy lint which would catch this...

I would normally write `Endian::from_str` for clarity, but I see your point

Cargo.toml
```diff
@@ -32,6 +32,7 @@ hex-literal = "0.3"
 rstest = "0.6"
 criterion = "0.3"
 alloc_counter = "0.0.4"
+trybuild = "1.0"
```
sharksforarms, line 35: never used `trybuild` before! Looks awesome.

sharksforarms:
> Ahhhhh so many things need to test. I even don't know what thing is acceptable what is not.

I do have coverage enabled once the tests pass (I think it's one .stderr file missing), may be easier to cross-ref with that to see what's being hit!

examples/custom_reader_and_writer.rs
```diff
@@ -1,3 +1,6 @@
+// NO CUSTOM READER/WRITER WHICH USING `input_is_le` and `bit_size`
```
sharksforarms, line 1: TODO: Fix example

sharksforarms:
Just a couple last comments. Great code and very appreciated contribution! Thanks!

tests/macro_read/unknown_attribute.stderr
```diff
@@ -0,0 +1,35 @@
...
+ --> $DIR/unknown_attribute.rs:4:8
+  |
+4 | #[deku(a)]
+  |        ^
+
+error: Unknown field: `a`
+ --> $DIR/unknown_attribute.rs:9:12
+  |
+9 |     #[deku(a = 1)]
+  |            ^^^^^
```
sharksforarms, line 11: latest rust stable does
```rust
9 |     #[deku(a = 1)]
  |                  ^
```

constfold: Yeah, but don't know why it was changed in nightly. I will use stable next time.

constfold:
What the hell is going on?
Okay, 
`CI / Test Suite (pull_request)` uses `stable-x86_64-unknown-linux-gnu - rustc 1.45.0 (5c1f21c3b 2020-07-13)` 
but 
`CI / Coverage (pull_request) ` uses `nightly-x86_64-unknown-linux-gnu - rustc 1.47.0-nightly (5ef299eb9 2020-07-24)`

constfold:
Aha, It's because `quote` use [`Span::join`](https://docs.rs/proc-macro2/1.0.19/proc_macro2/struct.Span.html#method.join) for span concat  ([src](https://github.com/dtolnay/quote/blob/1d757d4bda76933aaf64005b00bd0e43108c5cee/src/spanned.rs#L40)), but it's  nightly-only. Looks like we need to parse attribute without `darling` to make sure an invalid attribute will always generate compile error like
```
#[deku(aaa = 1)]
       ^^^
```
instead(Because the lack of `Span::join`, generate error like below is impossible.)
```
#[deku(aaa = 1)]
       ^^^^^^^
```

sharksforarms:
Ah that's a pain, I think the coverage only works on nighty... Might be possible to feature flag the test case to only run on stable...

I'll get back to this soon! Away for a couple of days.

sharksforarms:
Merged into master. Thank you very much for your time and energy towards this feature!

PR #64, opened 2020-07-27
Use pass by value (1 byte vs 8 bytes)

wcampbell0x2a:


sharksforarms:
Hey @wcampbell0x2a thanks for the PR! What are your intentions for the change? Performance? Code style?

This would make it consume ownership of the enum instead of borrowing it

wcampbell0x2a:
> Hey @wcampbell0x2a thanks for the PR! What are your intentions for the change? Performance? Code style?
> 
> This would make it consume ownership of the enum instead of borrowing it

It was mostly for performance, I didn't think the consuming would be a problem for the API.

sharksforarms:
I'm not sure the performance benefits (would be super minimal, if anything) outweigh the usefulness of the borrowing API, what do you think @constfold ? (Freshly merged code here)

constfold:
Looks good to me.
> This would make it consume ownership of the enum instead of borrowing it

It's okay, `Endian` is `Copy`.

> I'm not sure the performance benefits (would be super minimal, if anything) 

`Endian` might be 1 bit, but a ref is always usize(4 bytes or 8 bytes). Maybe it's useful, in few cases.

PR #66, opened 2020-07-28
Remove test module

wcampbell0x2a:
Module `tests` is unnecessary when #[test] is used in the tests
directory.


sharksforarms:
Thanks! :D

PR #73, opened 2020-07-28
Replace `Option<String>` with `Option<LitStr>`

### LINKED ISSUE #72: `option_as_tokenstream` will eat span
Since `option_as_tokenstream` uses `Option<String>` as its input, `darling` will discard the span while parsing(because `String` doesn't have a span). It makes error message hard to read.
For example:
```rust
#[derive(DekuRead)]
struct Foo {
    #[deku(cond = "'a' == 2")]
    a: u8,
}
```
error:
```rust
error[E0308]: mismatched types
  --> src\main.rs:24:10
   |
24 | #[derive(DekuRead)]
   |          ^^^^^^^^ expected `char`, found `u8`
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
replace it with `LitStr`:
```rust
error[E0308]: mismatched types
  --> src\main.rs:26:19
   |
26 |     #[deku(cond = "'a' == 2")]
   |                   ^^^^^^^^^^ expected `char`, found `u8`
```

constfold:
Fixes #72 
We should also add some tests

### DIFF
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -326,7 +326,7 @@ struct DekuReceiver {
 }
 
 /// Parse a TokenStream from an Option<String>
-fn option_as_tokenstream(input: Option<String>) -> Option<TokenStream> {
+fn option_as_tokenstream(input: Option<syn::LitStr>) -> Option<TokenStream> {
     input.map(|v| {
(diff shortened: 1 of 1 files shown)


sharksforarms:
Merged in master! Thanks

Issue #77, opened 2020-07-29
add top-level enum attribute `id`


wcampbell0x2a:
Currently I don't see a way to use ctx as the bytes to parse an enum (instead of reading bits/bytes). The key part is the `category` *and* `length` need to be read before the `Messages` are parsed. The `category` field is used to decide which struct from an enum is parsed.

### Example
```
#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
struct AsterixPacket {
    #[deku(bytes = "1", endian = "big")]
    category: u8,
    #[deku(bytes = "2", endian = "big")]
    length: u16,
    #[deku(ctx = "*category")]
    messages: Vec<Message>,
}

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
#[deku(id_ctx = "category")]
enum Message {
    #[deku(id = "48")]
    Cat48(Cat48),
}
```

sharksforarms:
So a way to "get the ID from somewhere instead of consuming"

wcampbell0x2a:
> So a way to "get the ID from somewhere instead of consuming"

Correct

sharksforarms:
Possibly with `id` on the top-level

```rust
#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
struct AsterixPacket {
    #[deku(bytes = "1", endian = "big")]
    category: u8,
    #[deku(bytes = "2", endian = "big")]
    length: u16,
    #[deku(ctx = "*category")]
    messages: Vec<Message>,
}

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
#[deku(ctx = "category: u8", id = "category")]
enum Message {
    #[deku(id = "48")]
    Cat48(Cat48),
}
```

sharksforarms:
Kinda got a working implementation in the works, stay tuned!

PR #78, opened 2020-07-30
Add `id` attribute to top-level enum

sharksforarms:
Fixes https://github.com/sharksforarms/deku/issues/77


sharksforarms:
@wcampbell0x2a take a look and let me know if it meets your use-case

wcampbell0x2a:
> @wcampbell0x2a take a look and let me know if it meets your use-case

Works beautifully

Issue #79, opened 2020-08-01
Endian-ness composing


wcampbell0x2a:
I believe it would be a nice features for child structs and enums to inherit the parents endian type.

Currently the following code produces the following error:
```
 `deku::DekuRead<deku::ctx::Endian>` is not implemented for B
```
```
use deku::prelude::*;

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
#[deku(endian = "big")]
struct Packet {
    len: u16,
    messages: B,
}

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
#[deku(id_type = "u8")]
enum B {
    #[deku(id = "0x00")]
    one,
    #[deku(id = "0x01")]
    two,
    #[deku(id = "0x02")]
    three,
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test01() {
        let data: Vec<u8> = vec![0x04, 0x13, 0x01];

        let (_, value) = Packet::from_bytes((data.as_ref(), 0)).unwrap();

        assert_eq!(
            Packet {
                len: 0x0413,
                messages: B::two,
            },
            value
        );
    }
}
```

In fact,  the way of creating a compiling version of this code seems a bit odd. As I only add endian to the `len` field.
```
use deku::prelude::*;

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
struct Packet {
    #[deku(endian = "big")]
    len: u16,
    messages: B,
}

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
#[deku(id_type = "u8")]
enum B {
    #[deku(id = "0x00")]
    one,
    #[deku(id = "0x01")]
    two,
    #[deku(id = "0x02")]
    three,
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test01() {
        let data: Vec<u8> = vec![0x04, 0x13, 0x01];

        let (_, value) = Packet::from_bytes((data.as_ref(), 0)).unwrap();

        assert_eq!(
            Packet {
                len: 0x0413,
                messages: B::two,
            },
            value
        );
    }
}
```

constfold:
> child structs and enums to inherit the parents endian type.

Currently they do inherit. The error you got means we assume that `B` requires a `Endian` as its context, but it doesn't.(We can't imply implement that for `B` because we don't know if it need an endian or not, and how to use it.)
You need to add the requirement of `Endian` manually:
```rust
[...]

PR #80, opened 2020-08-01
doc: Update documentation surrounding `ctx`

sharksforarms:
In support of: https://github.com/sharksforarms/deku/issues/79


sharksforarms:
Generate using `cargo doc --workspace`

Issue #89, opened 2020-08-19
Cleanup enum attribute names


sharksforarms:
I can't think of a reason why there's both `id_bits` and `bits`, for example there's `endian` but not `id_endian` (we use `endian` for enum)

This would be more consistent with field/structs

Also rename `id_type` to `type` and `id` to `value`

Before:

```rust
#[deku(id_type = "u8", id_bits = "5")]
enum Test {
    #[deku(id = "0x01")]
    VarA,
}
```

After
```rust
#[deku(type = "u8", bits = "5")]
enum Test {
    #[deku(value = "0x01")]
    VarA,
}
```

wcampbell0x2a:
Would we also want `id_pat` changed to `value_pat`?

sharksforarms:
Thanks @wcampbell0x2a for the PRs!

> 
> 
> Would we also want `id_pat` changed to `value_pat`?

Hm, I'm starting to think `id` might be an _OK_ name for now?

wcampbell0x2a:
Personally, I think `type`, `id`, `id_pat` is fine.

If it were to go with `match`, I would think maybe `matches` might suit better since it fits better with the rust-std `matches` then the rust-std `match` syntax.

Naming is hard.

PR #94, opened 2020-09-02
Deprecate `id_type` in favor of `type`

wcampbell0x2a:
- Change darling input and use the darling rename attribute to take in `type`
  instead of `id_type`. This keeps the use of `id_type` internally,
  because of type being a rust-lang protected word.
- Update error messages, test fields

See #89


sharksforarms:
Thanks! Re-based it into master.

sharksforarms:

PR #95, opened 2020-09-02
Deprecate `id_bits` in favor of `bits`

wcampbell0x2a:
- More direct change, with a full change of all mentions of `id_bits`
  to `bits`.

See #89


sharksforarms:
Thanks!

sharksforarms:

PR #101, opened 2020-09-25
Change DekuWrite to accept a &mut BitVec instead of allocating

### LINKED ISSUE #59: DekuWrite write() should take &mut BitVec
Instead of having an allocation per field, write should take a mut bitvec by ref to extend.

agausmann:
Resolves #59 

**Breaking change:**

- This breaks the interface required by the `DekuWrite` trait, since `DekuWrite::write` accepts a `&mut BitVec` instead of returning one. I think this is justified since it is an objectively better API and since this crate is not yet 1.0, some API churn should be expected.

- There are some differences at runtime regarding the layout of the output `BitVec`s, requiring changes to some test cases ([explained in comments below](https://github.com/sharksforarms/deku/pull/101#discussion_r494705816)).

Todo:

- [x] Update derive macros to output correct impls for the new API
- [x] Update integration/macro tests
- [x] Update doctests
- [x] Update examples

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -419,53 +419,57 @@ macro_rules! ImplDekuTraits {
 
                 let input_bits = input.view_bits::<Msb0>();
 
-                    if bit_size > input_bits.len() {
-                        return Err(DekuError::InvalidParam(format!(
-                            "bit size {} is larger then input {}",
-                            bit_size,
-                            input_bits.len()
-                        )));
-                    }
+                if bit_size > input_bits.len() {
+                    return Err(DekuError::InvalidParam(format!(
+                        "bit size {} is larger then input {}",
+                        bit_size,
+                        input_bits.len()
+                    )));
+                }
 
-                    if output_is_le {
-                        // Example read 10 bits u32 [0xAB, 0b11_000000]
-                        // => [10101011, 00000011, 00000000, 00000000]
-                        let mut remaining_bits = bit_size;
-                        for chunk in input_bits.chunks(8) {
-                            if chunk.len() > remaining_bits {
-                                let bits = &chunk[chunk.len() - remaining_bits..];
-                                for b in bits {
-                                    output.push(*b);
-                                }
-                                // https://github.com/myrrlyn/bitvec/issues/62
-                                // output.extend_from_slice(chunk[chunk.len() - remaining_bits..]);
-                                break;
-                            } else {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -813,16 +806,17 @@ mod tests {
         case::normal_le(0xDDCC_BBAA, Endian::Little, None, vec![0xAA, 0xBB, 0xCC, 0xDD]),
         case::normal_be(0xDDCC_BBAA, Endian::Big, None, vec![0xDD, 0xCC, 0xBB, 0xAA]),
         case::bit_size_le_smaller(0x03AB, Endian::Little, Some(10), vec![0xAB, 0b11_000000]),
-        case::bit_size_be_smaller(0x03AB, Endian::Big, Some(10), vec![0b11, 0xAB]),
+        case::bit_size_be_smaller(0x03AB, Endian::Big, Some(10), vec![0b11_1010_10, 0b11_000000]),
```
agausmann, line 809: I had to change the layout of this output vec because of the alignment of the bits written to the output `BitVec`. Originally, the `BitVec` returned by `DekuWrite::write` was created directly from the array of bytes created from an integer (in this case, `0x03AB.to_be_bytes()  == [0b000000_11, 0b1010_1011]`), with an offset pointing to the first byte of the output.

With the new API, an empty `BitVec` is created by the caller and passed in, where the bytes are written MSB-first, so that the backing array is `[0b11_1010_10, 0b11_000000]` - the previous array, but shifted left by 6 bits.

When `BitVec::into_vec` is called in the test harness, it reuses the backing array without shifting or re-aligning the data it contains, so the offset to the first bit is unchanged. Originally, this meant 6 bits of left-padding, which is now gone.

sharksforarms:
Thanks for tackling this @agausmann, it'll be interesting to see the impact! I'll see if I can add more benchmarks to master to compare.

agausmann:
Benchmark results aren't very shocking (around -50%), because they're still dominated by the allocation made by `DekuContainerWrite::to_bytes()`. You should see much more significant improvements at scale.

sharksforarms:
Looks good! If you could rebase on master and resolve that one conflict that'd be appreciated!

As for the test case, it makes sense to me. Thanks for explaining.

sharksforarms (approved):
Thanks! Commits are in master

sharksforarms:

PR #102, opened 2020-09-27
Use match and matches for checking endian

wcampbell0x2a:
This is cleaner imo and less likely for errors by using this pattern
instead of `if`s.


sharksforarms:
Thanks!

PR #106, opened 2020-10-14
Check container size vs bits/bytes size at compile

wcampbell0x2a:
- Add panic if bits is larger than the container size.
- Add Tests for new feature.
- Removed code in runtime for checking container size, since it is
  asserted at compile time now.

Any comments/questions welcome!

### DIFF
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -591,10 +591,26 @@ mod tests {
             #[deku(id = "3")]
             C { field_n: u8 },
         }"#),
+        case::valid_bits(r#"struct Test(#[deku(bits=8)] u8);"#),
 
-        // TODO: these tests should error/warn eventually?
-        // error: trying to store 9 bits in 8 bit type
+        // error: invalid bit container size for deku bits/bytes attribute
+        #[should_panic(expected="field `field_0` of type `u8` with bit size: 8, not big enough for bits = 9")]
         case::invalid_storage(r#"struct Test(#[deku(bits=9)] u8);"#),
+        #[should_panic(expected="field `field_a` of type `u16` with bit size: 16, not big enough for bits = 24")]
+        case::invalid_storage(r#"
+        struct Invalid {
+            #[deku(bytes = 3)]
+            pub field_a: u16
+        }
+        "#),
+        #[should_panic(expected="field `ganondorf` of type `u8` with bit size: 8, not big enough for bits = 120")]
+        case::invalid_storage(r#"
+        struct Invalid {
+            #[deku(bits = 120)]
+            pub ganondorf: u8
+        }
+        "#),
+        // TODO
         // warn: trying to set endian on a type which wouldn't make a difference
         case::invalid_endian(r#"struct Test(#[endian=big] u8);"#),
     )]
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -616,6 +616,13 @@ mod tests {
             pub ganondorf: u8
         }
         "#),
+        #[should_panic(expected="field `ganondorf` of type `u8` with bit size: 8, not big enough for bits = 9")]
+        case::invalid_storage(r#"
+        struct Invalid {
+            #[deku(ctx = "deku::ctx::BitSize(9)")]
+            pub ganondorf: u8
+        }
+        "#),
         // TODO
         // warn: trying to set endian on a type which wouldn't make a difference
         case::invalid_endian(r#"struct Test(#[endian=big] u8);"#),
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -578,6 +578,12 @@ mod tests {
             #[deku(skip, default = "5")]
             field_e: u32,
         }"#),
+        case::valid_ctx_bitsize(r#"
+        struct Invalid {
+            #[deku(ctx = "deku::ctx::BitSize(7)")]
+            pub hey_listen: u8
+        }
+        "#),
 
         // Valid Enum
         case::enum_empty(r#"#[deku(type = "u8")] enum Test {}"#),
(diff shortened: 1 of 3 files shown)


deku-derive/src/lib.rs
```diff
@@ -591,10 +591,26 @@ mod tests {
...
+        #[should_panic(expected="field `field_0` of type `u8` with bit size: 8, not big enough for bits = 9")]
         case::invalid_storage(r#"struct Test(#[deku(bits=9)] u8);"#),
+        #[should_panic(expected="field `field_a` of type `u16` with bit size: 16, not big enough for bits = 24")]
+        case::invalid_storage(r#"
+        struct Invalid {
+            #[deku(bytes = 3)]
+            pub field_a: u16
+        }
+        "#),
+        #[should_panic(expected="field `ganondorf` of type `u8` with bit size: 8, not big enough for bits = 120")]
```
sharksforarms, line 606: Heyyy someone got the deku reference! :+1:

src/lib.rs
```diff
@@ -288,13 +288,6 @@ macro_rules! ImplDekuTraits {
 
                 let input_is_le = endian.is_le();
 
-                if bit_size > max_type_bits {
```
sharksforarms, line 291: This is great! Unfortunately, there's also this case...

```rust
struct DekuTest {
    #[deku(ctx = "deku::ctx::BitSize(9)")]
    field_a: u8
}
```

wcampbell0x2a: Should be fixed now

deku-derive/src/macros/deku_read.rs
```diff
@@ -449,11 +467,35 @@ fn check_bitsize_container(
...
-                        field_ident, ident, max_bit_size, bit_size
-                    );
+                check_bitsize(max_bit_size, type_bit_size, field_ident, type_ident);
+            }
+        }
+    }
+}
+
+/// Convert `ctx::BitSize` into usize and check with `check_bitsize_bits_bytes`
+fn check_bitsize_ctx(
```
sharksforarms, line 477: This is clever (and probably still useful to have!), however because it's an expression, we can't assume it's contents. I think a dynamic check is still required.

```rust
struct DekuTest {
    #[deku(ctx = "deku::ctx::BitSize(std::mem::size_of::<u16>())")]
    field_a: u8
}
```

wcampbell0x2a: Didn't think about that, how would I check that dynamically at compile time?

sharksforarms: That's the main issue, I'm not sure of how this can be done. It's almost like code needs to be generated, then compiled & executed as opposed to just executed in the proc-macro

sharksforarms:
@wcampbell0x2a I think I found a solution.

Basically, we need to create a `const fn`, add the checking code, and call it.

Here's some sample code. `panic!` is not `const fn`, so I hacked around an assertion to test. But this seems to work!
```rust
    let bits_check = if let Some(bits) = f.bits {
        quote! {
            const fn __tester() {
                pub const ASSERT: [(); 1] = [()];
                if #bits > core::mem::size_of::<#field_type>() * 8 {
                    ASSERT[100];
                }
            }
            const a: () = __tester();
        }
    } else {
        quote!()
    };
```

There's different ways to implement this, we could either add a `Vec<TokenStream>` to the `emit_reads` function, or we could create an `emit_field_verification` ? Something along those lines possibly

sharksforarms:
Hm, not sure how we'd get a better error message than "index out of bounds"

sharksforarms:
Alternatively, we can add the runtime check if the validation functions you added failed (only add extra runtime cost for the off case)

constfold (commented):
We can use [static_assertion](https://docs.rs/static_assertions) to do those works. But I still don't think there is a way to check the size in compile time since we allow every expressions in `ctx` attribute. We need to such a thing like eval it at compile time if the expr is const or eval it in runtime if it's not.

wcampbell0x2a:
Just wondering about general thoughts on this issue.

I like the const fn approach, but where does [staic_assertions](https://docs.rs/static_assertions) fit into this?

Just trying to figure out what the best way to support the `ctx` stuff

sharksforarms:
> I like the const fn approach, but where does [static_assertions](https://docs.rs/static_assertions) fit into this?

It's a crate which does the `const fn` example above in a better/similar way :)
See https://docs.rs/static_assertions/1.1.0/src/static_assertions/const_assert.rs.html#55

The con of this approach is the lack of feedback to the user... it looks like this just panics with an index out of bounds? https://github.com/nvzqz/static-assertions-rs/issues/7

inspier:
Doing something like this might be able to provide better user feedback: [Multiple of 8 const assertion](https://github.com/dtolnay/case-studies/blob/master/bitfield-assertion/README.md).

sharksforarms:
Going to close this for now

sharksforarms:
> Alternatively, we can add the runtime check if the validation functions you added failed (only add extra runtime cost for the off case)
> 
> This is probably better until panic! is [available inside const fn](https://github.com/rust-lang/rust/issues/51999)

@wcampbell0x2a this is now the case in rust 1.57.0 : https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html

constfold:
> this is now the case in rust 1.57.0
Finally! I love compile time checks.

wcampbell0x2a:
> > Alternatively, we can add the runtime check if the validation functions you added failed (only add extra runtime cost for the off case)
> > This is probably better until panic! is [available inside const fn](https://github.com/rust-lang/rust/issues/51999)
> 
> @wcampbell0x2a this is now the case in rust 1.57.0 : https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html

sweet

Issue #109, opened 2020-10-20
Support magic values prefixing types


samuelsleight:
It would be nice to support an attributes similar to `binread`'s `magic` attribute, for specifying a type must start with a given byte sequence, like so:

```rust
#[derive(Debug, DekuRead)]
#[deku(magic = b"ID3")]
pub struct Id3v2 {
   ...
}
```

This is something I have hacked in on a branch ([here](https://github.com/samuelsleight/deku/commit/507ecfccc66b1e10fa53ed7daf168fdda23498f8)), however I figured it would be sensible to open a discussion about it and actually do it properly (and support writing in addition to reading).

wcampbell0x2a:
Somewhat related: https://github.com/sharksforarms/deku/issues/71

samuelsleight:
Ah, I'd misssed that that issue was covering magic values also. 
Potentially the best way of implementing this feature (or at least the reading half of it) is using a hidden `assert_eq` under-the-hood

sharksforarms:
Great idea! Are you interested in submitting a PR with your branch? If not, I can take a look at your branch this week

samuelsleight:
My branch as-is is just something I hacked in to get something working when porting my code from binread, and I'm not necessarily sold on it being the _best_ way of doing things. If you think a PR is a better place to work out what _is_ the best way then I'm happy to open one now, though?

sharksforarms:
No problem! I was mainly trying to figure out intention ;) I appreciate all the maintenance/contribution help I can get! If you wanted to open a PR, I can guide you through the code/adding tests/adding a write impl/docs

samuelsleight:
In that case, I may as well open one then, why not

PR #112, opened 2020-10-20
Support magic values prefixing types

samuelsleight:
This PR as-is is an initial implementation of #109, hacked-in with very little experience in this library or with developing proc-macros in general.

It currently supports reads only, and is very specific to my use-case, so will want to be expanded to support writes and to be more useful in general, and will also want tests implementing and documentation adding for the new attribute.

### DIFF
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -29,6 +29,9 @@ struct DekuData {
     /// default context passed to the field
     ctx_default: Option<Punctuated<syn::Expr, syn::token::Comma>>,
 
+    /// A magic value that must appear at the start of this struct's data
+    magic: Option<syn::LitByteStr>,
+
     /// enum only: `id` value
     id: Option<TokenStream>,
 
@@ -87,6 +90,7 @@ impl DekuData {
             endian: receiver.endian,
             ctx,
             ctx_default,
+            magic: receiver.magic,
             id: receiver.id,
             id_type: receiver.id_type,
             bits,
@@ -385,6 +389,10 @@ struct DekuReceiver {
     //       https://github.com/TedDriggs/darling/pull/98
     #[darling(default)]
     ctx_default: Option<syn::LitStr>,
+
+    /// A magic value that must appear at the start of this struct's data
+    #[darling(default)]
+    magic: Option<syn::LitByteStr>,
 
     /// enum only: `id` value
     #[darling(default, map = "option_as_tokenstream")]
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -29,7 +29,7 @@ struct DekuData {
     /// default context passed to the field
     ctx_default: Option<Punctuated<syn::Expr, syn::token::Comma>>,
 
-    /// A magic value that must appear at the start of this struct's data
+    /// A magic value that must appear at the start of this struct/enum's data
     magic: Option<syn::LitByteStr>,
 
     /// enum only: `id` value
@@ -390,7 +390,7 @@ struct DekuReceiver {
     #[darling(default)]
     ctx_default: Option<syn::LitStr>,
 
-    /// A magic value that must appear at the start of this struct's data
+    /// A magic value that must appear at the start of this struct/enum's data
     #[darling(default)]
     magic: Option<syn::LitByteStr>,
 
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -32,28 +32,13 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
         .and_then(|v| v.ident.as_ref())
         .is_some();
 
+    let magic_read = emit_magic_read(input)?;
+
     let (field_idents, field_reads) = emit_field_reads(input, &fields)?;
 
     let internal_fields = gen_internal_field_idents(is_named_struct, field_idents);
 
     let initialize_struct = super::gen_struct_init(is_named_struct, internal_fields);
-
-    let check_magic = if let Some(magic) = &input.magic {
-        quote!{
-            let magic = #magic;
-
-            for byte in magic {
-                let (new_rest, read_byte) = u8::read(rest, ())?;
-                if *byte != read_byte {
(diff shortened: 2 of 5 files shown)


deku-derive/src/macros/deku_read.rs
```diff
@@ -38,6 +38,23 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
 
     let initialize_struct = super::gen_struct_init(is_named_struct, internal_fields);
 
+    let check_magic = if let Some(magic) = &input.magic {
```
sharksforarms, line 41: Similarly, check_magic should also be available for [emit_enum](https://github.com/sharksforarms/deku/blob/22e271d421574cf9ebc6f9e00c42feb0978ecb70/deku-derive/src/macros/deku_read.rs#L249)

deku-derive/src/lib.rs
```diff
@@ -29,6 +29,9 @@ struct DekuData {
     /// default context passed to the field
     ctx_default: Option<Punctuated<syn::Expr, syn::token::Comma>>,
 
+    /// A magic value that must appear at the start of this struct's data
```
sharksforarms, line 32: `A magic value that must appear at the start` (for both structs/enums)

deku-derive/src/lib.rs
```diff
@@ -386,6 +390,10 @@ struct DekuReceiver {
     #[darling(default)]
     ctx_default: Option<syn::LitStr>,
 
+    /// A magic value that must appear at the start of this struct's data
```
sharksforarms, line 393: `A magic value that must appear at the start` (for both structs/enums)

sharksforarms (commented):
Looks good so far! Couple of comments
- A function could be created such as `emit_magic` in the same file to share the "magic reading" tokens between `emit_struct` and `emit_enum`
- A success/failure test case should be added here: https://github.com/sharksforarms/deku/tree/master/tests/test_attributes
- Documentation of the new attribute should be added here: https://github.com/sharksforarms/deku/blob/master/src/attributes.rs#L6
- deku_write.rs write implementation, layout is similar to deku_read.rs (emit_struct/emit_enum) (write magic value before writing fields): https://github.com/sharksforarms/deku/blob/master/deku-derive/src/macros/deku_write.rs

samuelsleight:
Thanks for the pointers! I'll hopefully get a chance to look into these tomorrow

samuelsleight:
Cool, this should now be somewhat less hacked-in!

src/attributes.rs
```diff
@@ -101,6 +102,36 @@ let value: Vec<u8> = value.try_into().unwrap();
...
+that type's data when writing, without needing to be explicitly stored
+in a field
+
+Example:
+```rust
+# use deku::prelude::*;
+# use std::convert::{TryInto, TryFrom};
+# #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+#[deku(magic = b"deku")]
+pub struct MagicDeku {
```
sharksforarms, line 118: nit: `DekuTest` is used as the struct names in the docs for consistency between examples

tests/test_magic.rs
```diff
@@ -0,0 +1,105 @@
+use deku::prelude::*;
+use hexlit::hex;
+use rstest::rstest;
+use std::convert::{TryFrom, TryInto};
+
+pub mod samples {
```
sharksforarms, line 6: I got rid of this `samples` pattern during a recent refactor, you can place the structs/enums inside the test functions :) Can check `test_cond.rs` for an example

tests/test_magic.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[deku(magic = b"UKED")]
+    pub struct NestedMagicDeku {
+        pub nested: MagicDeku,
+    }
+}
+
+#[rstest(input,
+    case(&hex!("64656b75")),
+
+    #[should_panic(expected = "Parse(\"Missing magic value [100, 101, 107, 117]\")")]
```
sharksforarms, line 30: Are the multiple "Missing magic value" tests testing different code paths?

tests/test_magic.rs
sharksforarms: Also used `TestStruct` or `TestEnum` inside the test functions for consistency

tests/test_magic.rs
```diff
@@ -0,0 +1,105 @@
...
+#[rstest(input,
+    case(&hex!("554b454464656b75")),
+
+    #[should_panic(expected = "Parse(\"Missing magic value [100, 101, 107, 117]\")")]
+    case(&hex!("554b4544deadbeef")),
+
+    #[should_panic(expected = "Parse(\"Missing magic value [85, 75, 69, 68]\")")]
+    case(&hex!("deadbeef64656b75")),
+)]
+fn test_nested_magic_struct(input: &[u8]) {
```
sharksforarms, line 93: I don't think this test case is necessary, there's a test case for nested cases in `test_struct.rs`

src/attributes.rs
```diff
@@ -101,6 +102,36 @@ let value: Vec<u8> = value.try_into().unwrap();
 assert_eq!(data, value);
 ```
 
+# magic
+
+Sets a "magic" value that must be present in the data at the start of
+a struct/enum when reading, and that is written out of the start of
+that type's data when writing, without needing to be explicitly stored
```
sharksforarms, line 109: `, without needing to be explicitly stored in a field` can be removed I think

sharksforarms (commented):
Looks great! Some small comments.

tests/test_magic.rs
samuelsleight: Oh, yep, I missed that that refactor had removed the `samples` pattern

tests/test_magic.rs
samuelsleight: They are each testing a different byte of the input being incorrect, which I figured was worth checking for to future-proof against really bad implementations of magic reading that, say, only check for the last byte and assume the first few are correct.

(Not saying that specific case is going to happen, obviously, but I just figured it was better to be safe than sorry)

sharksforarms:
Thanks!

PR #114, opened 2020-10-20
Avoid `use core::convert::TryInto` in empty update

wcampbell0x2a:
This is rather small, but it removes the quotes! for:
`use core::convert::TryInto` if the update() function has no need for
it.

### DIFF
--- a/deku-derive/src/macros/deku_write.rs
+++ b/deku-derive/src/macros/deku_write.rs
@@ -98,10 +98,13 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
         }
     };
 
+    // avoid outputing `use core::convert::TryInto` if update() function is empty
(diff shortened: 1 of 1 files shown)


deku-derive/src/macros/deku_write.rs
```diff
@@ -98,10 +98,13 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
         }
     };
 
+    // avoid outputing `use core::convert::TryInto` if update() function is empty
+    let update_use = check_use_update(&field_updates);
```
sharksforarms, line 102: `check_update_use` to stay consistent with variable name?

wcampbell0x2a: :+1:  done

sharksforarms:
Thanks!

PR #115, opened 2020-10-20
Add emit_try_from() and emit_from_bytes()

wcampbell0x2a:
Add functions for shared usage between enum and structs emit functions for removal of duplicated code.


sharksforarms:
Thanks!

Issue #110, opened 2020-10-20
Read a container until a given condition is met


samuelsleight:
There are a couple of use-cases for this: FLAC metadata is read until a given flag in the metadata header is set marking it as the last one, and also for reading null-terminated strings. In both these cases I have manually used a `reader` attribute to read in a loop until said condition holds, however it might be nice to support an `until` attribute on vectors that takes a predicate to call on each item as it is read, for example:

```rust
#[derive(DekuRead)]
struct NullTerminatedString {
    #[deku(until = "|byte| byte == 0")]
    vec: Vec<u8>,
}
```

sharksforarms:
Agreed, this would be useful and is both read/write compatible.

samuelsleight:
So, I've been thinking about this one and I'm leaning towards implementing this by converting the current `Count` context type into an `enum` of either `Count` or `Until` - this feels like a nice clean way of enforcing them being mutually exclusive (which I think they should be) and also doesn't require any new `impl` blocks, we can just modify the current blocks that take `Count`. It's also then nicely extensible with any future ways of reading a `Vec`. It would, however be a breaking change for anybody directly using `Count`, which may or may not be a good idea to introduce.
Do we think this is a good idea?

sharksforarms:
Possibly related: https://github.com/sharksforarms/deku/issues/76

sharksforarms:
> So, I've been thinking about this one and I'm leaning towards implementing this by converting the current `Count` context type into an `enum` of either `Count` or `Until` - this feels like a nice clean way of enforcing them being mutually exclusive (which I think they should be) and also doesn't require any new `impl` blocks, we can just modify the current blocks that take `Count`. It's also then nicely extensible with any future ways of reading a `Vec`. It would, however be a breaking change for anybody directly using `Count`, which may or may not be a good idea to introduce.

Breaking changes are OK with me. This library is still changing and is not committed to stability (1.0) just yet.

> Do we think this is a good idea?

Yes! I think that's a good way of handling mutual exclusivity between different "contexts"

Issue #76, opened 2020-07-29
Attribute to handle length of bytes(or bits) in Vec<T> with other struct field


wcampbell0x2a:
Consider the following code:
```
use deku::prelude::*;

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
pub struct Packet {
    #[deku(bytes = "1")]
    length: u8,
    // byte len of all of messages is length - 2
    messages: Vec<Message>,
}

/// In the real packet, this would be of variable length, so we can't use just the `count` attribute on messages
#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
pub struct Message {
    #[deku(bytes = "1")]
    msg: u8,
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test01() {
        let data: Vec<u8> = vec![0x04, 0x01, 0x02];

        let (_, value) = Packet::from_bytes((data.as_ref(), 0)).unwrap();

        assert_eq!(
            Packet {
                length: 0x04,
                messages: vec![Message { msg: 0x01 }, Message { msg: 0x02 }]
            },
            value
        );
    }
}
```

I can use the `count` attribute to give the length of `Vec<T>` elements, but I see no way of telling the max bytes that a Vec<T> can have in it's container as a whole. Wondering if this would be a feature, or do I need to do some weird custom write/read implementation with a `read_bytes` field.

sharksforarms:
Just to be clear, `#[deku(count = "length - 2")]` on `messages` doesn't satisfy this?

wcampbell0x2a:
> Just to be clear, `#[deku(count = "length - 2")]` on `messages` doesn't satisfy this?

Correct, since that is the length of elements in the Vec<T>.

constfold:
Does `ctx` satisfy this situation?
```rust
struct A {
    len: u8,
    #[deku(ctx = "len")]
    messages: Vec<Message>
}
#[deku(ctx = "byte_len: u8")]
struct Message {
    #[deku(bytes = "byte_len")]
    msg: u8,
}
```

wcampbell0x2a:
I don't think so, since it's not the length of the bytes in one Message, it's the length of *All* the Messages in a Vec.

Reminder that 

However, having the following issues when testing your code:
```
use deku::prelude::*;

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
pub struct Packet {
    len: u8,
    // byte len of all of messages is length - 2
[...]

PR #118, opened 2020-10-22
Implement the backend for reading a vector until a predicate or until a number of bits

samuelsleight:
This PR implements a potential backend (i.e., the context types only, and no proc-macros to use them yet) for both #110 and #76. I've done it like this to make sure we're happy with how this is implemented before I go any further, as this is a rather important breaking change (and also because the latter of those issues has not yet reached a conclusion as to the attribute it would use).

We convert the `Count` context type into a `Limit` enum, allowing either a count (the original behaviour), a number of bits, or a general predicate.

Annoyingly, the `Limit` type is now quite ugly, due to the need to have type parameters for the `Until` variant - I cannot think of any particularly nice options here. The best alternative is probably to template `Limit` on `<T>` only, and then take a `&mut dyn FnMut` in the `Until` variant, but I'm open to any suggestions here. Potentialy however, this ugliness doesn't _really_ matter, as the type is mostly designed to be hidden behind the nice proc-macro attributes once those are actually implemented for the two new cases.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -498,7 +498,7 @@ macro_rules! ImplDekuTraits {
     };
 }
 
-fn read_vec_with_predicate<T: DekuRead<Ctx>, Ctx: Copy, Predicate: FnMut(&T) -> bool>(
+fn read_vec_with_predicate<T: DekuRead<Ctx>, Ctx: Copy, Predicate: FnMut(usize, &T) -> bool>(
     input: &BitSlice<Msb0, u8>,
     vec: Vec<T>,
     ctx: Ctx,
@@ -511,7 +511,7 @@ fn read_vec_with_predicate<T: DekuRead<Ctx>, Ctx: Copy, Predicate: FnMut(&T) -> 
         res.push(val);
         rest = new_rest;
 
-        if predicate(res.last().unwrap()) {
+        if predicate(input.offset_from(rest) as usize, res.last().unwrap()) {
             break;
         }
     }
@@ -551,15 +551,24 @@ impl<T: DekuRead<Ctx>, Ctx: Copy, Predicate: FnMut(&T) -> bool> DekuRead<(Limit<
                 }
 
                 // Otherwise, read until we have read `count` elements
-                read_vec_with_predicate(input, Vec::with_capacity(count), inner_ctx, move |_| {
+                read_vec_with_predicate(input, Vec::with_capacity(count), inner_ctx, move |_, _| {
                     count -= 1;
                     count == 0
                 })
             }
 
             // Read until a given predicate returns true
-            Limit::Until(predicate, _) => {
-                read_vec_with_predicate(input, Vec::new(), inner_ctx, predicate)
+            Limit::Until(mut predicate, _) => {
+                read_vec_with_predicate(input, Vec::new(), inner_ctx, move |_, value| {
+                    predicate(value)
+                })
+            }
+
+            // Read until a given quanity of bits have been read
+            Limit::Bits(bits) => {
+                read_vec_with_predicate(input, Vec::new(), inner_ctx, move |read_bits, _| {
+                    read_bits == bits.into()
+                })
             }
         }
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -896,25 +896,27 @@ mod tests {
         assert_eq!(expected_write, res_write.into_vec());
     }
 
-    #[rstest(input,endian,bit_size,count,expected,expected_rest,
-        case::count_0([0xAA].as_ref(), Endian::Little, Some(8), 0, vec![], bits![Msb0, u8; 1, 0, 1, 0, 1, 0, 1, 0]),
-        case::count_1([0xAA, 0xBB].as_ref(), Endian::Little, Some(8), 1, vec![0xAA], bits![Msb0, u8; 1, 0, 1, 1, 1, 0, 1, 1]),
-        case::count_2([0xAA, 0xBB, 0xCC].as_ref(), Endian::Little, Some(8), 2, vec![0xAA, 0xBB], bits![Msb0, u8; 1, 1, 0, 0, 1, 1, 0, 0]),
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -498,9 +498,32 @@ macro_rules! ImplDekuTraits {
     };
 }
 
-impl<T: DekuRead<Ctx>, Ctx: Copy> DekuRead<(Count, Ctx)> for Vec<T> {
-    /// Read the specified number of `T`s from input.
-    /// * `count` - the number of `T`s you want to read.
+fn read_vec_with_predicate<T: DekuRead<Ctx>, Ctx: Copy, Predicate: FnMut(usize, &T) -> bool>(
+    input: &BitSlice<Msb0, u8>,
+    vec: Vec<T>,
```
sharksforarms, line 503: nit: possibly change to `&mut Vec<T>`

src/lib.rs
```diff
@@ -498,9 +498,32 @@ macro_rules! ImplDekuTraits {
...
+    mut predicate: Predicate,
+) -> Result<(&BitSlice<Msb0, u8>, Vec<T>), DekuError> {
+    let mut res = vec;
+    let mut rest = input;
+    loop {
+        let (new_rest, val) = <T>::read(rest, ctx)?;
+        res.push(val);
+        rest = new_rest;
+
+        if predicate(input.offset_from(rest) as usize, res.last().unwrap()) {
```
sharksforarms, line 514: Handle unwrap

src/lib.rs
```diff
@@ -498,9 +498,32 @@ macro_rules! ImplDekuTraits {
     };
 }
 
-impl<T: DekuRead<Ctx>, Ctx: Copy> DekuRead<(Count, Ctx)> for Vec<T> {
-    /// Read the specified number of `T`s from input.
-    /// * `count` - the number of `T`s you want to read.
+fn read_vec_with_predicate<T: DekuRead<Ctx>, Ctx: Copy, Predicate: FnMut(usize, &T) -> bool>(
```
sharksforarms, line 501: nit: Document function & include documentation on predicate arguments

src/lib.rs
```diff
@@ -514,35 +537,53 @@ impl<T: DekuRead<Ctx>, Ctx: Copy> DekuRead<(Count, Ctx)> for Vec<T> {
...
-        let count: usize = count.into();
-
-        let mut res = Vec::with_capacity(count);
-        let mut rest = input;
-        for _i in 0..count {
-            let (new_rest, val) = <T>::read(rest, inner_ctx)?;
-            res.push(val);
-            rest = new_rest;
-        }
+        match limit {
```
sharksforarms, line 545: I like this. Easy to read, clean and extensible.

sharksforarms (commented):
Looks great!

Question:
Is the Limit/predicate symmetrical? Missing impl DekuWrite for Vec<T>

samuelsleight:
> Is the Limit/predicate symmetrical? Missing impl DekuWrite for Vec

The intention is for this to function the same way `Count` previously did and only effect the reading, and indeed we never previously had any implementation of `DekuWrite` that was taking the previous `Count` and I believe there is no need for me to implement any for the other two cases I have added.

I should definitely add some tests to `test_vec_read_write` (which I missed when initially implementing this) to verify that that is indeed working as I expect, however

src/lib.rs
samuelsleight: I'm unsure what the best practice is for handling the unwrap here, given that it's trivially impossible for it to fail as we're pushing to the vec immediately before it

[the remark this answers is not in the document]
samuelsleight: As per the last comment on that issue is this now fine on `0.19.4`?

src/lib.rs
samuelsleight: Thinking about it, perhaps it's actually nicer to just take an optional capacity instead of the vec directly, and construct the vec inside this method? That feels slightly less weird to me than passing in the vec did, at least

src/lib.rs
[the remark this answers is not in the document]
sharksforarms: Yes, it's now fixed. You can keep it as is!

sharksforarms: Fair enough! :)
A quick comment saying that would help in the future when looking at possible panics

sharksforarms:
> > Is the Limit/predicate symmetrical? Missing impl DekuWrite for Vec
> 
> The intention is for this to function the same way `Count` previously did and only effect the reading, and indeed we never previously had any implementation of `DekuWrite` that was taking the previous `Count` and I believe there is no need for me to implement any for the other two cases I have added.
> 
> I should definitely add some tests to `test_vec_read_write` (which I missed when initially implementing this) to verify that that is indeed working as I expect, however

Sounds good.

Write can be implemented in other tickets as required.

Such as:
https://github.com/sharksforarms/deku/issues/120
https://github.com/sharksforarms/deku/issues/76

sharksforarms (approved):

sharksforarms:
@samuelsleight Just wanted to confirm you're done with the changes before merging! Thanks for your contribution!

samuelsleight:
> @samuelsleight Just wanted to confirm you're done with the changes before merging! Thanks for your contribution!

Yep! All done with this stage - the only question would be whether you wanted to wait for me to do the follow-up PR (which I've only partially implemented so far) before merging, otherwise it's just introducing a breaking change for minimal benefit

sharksforarms:
> > @samuelsleight Just wanted to confirm you're done with the changes before merging! Thanks for your contribution!
> 
> Yep! All done with this stage - the only question would be whether you wanted to wait for me to do the follow-up PR (which I've only partially implemented so far) before merging, otherwise it's just introducing a breaking change for minimal benefit

Yes, I don't plan on pushing a new version just yet. Most likely when this feature becomes usable!

samuelsleight:
> Yes, I don't plan on pushing a new version to crates.io just yet. Merging to master is ok. A release will be pushed when this feature becomes usable!

Cool, well, in that case, merge away! I should have a PR for the follow-up in the next couple of days :+1:

Issue #111, opened 2020-10-20
Expose the read byte offset as an available variable


samuelsleight:
(Apologies for the train of issues, this is the last one)

I have found it desirable to know how many bytes have been read. As an example, I am currently reading an id3v2 attached picture field like so:

```rust
#[derive(DekuRead)]
#[deku(ctx = "size: u32")]
pub struct Picture {
    encoding: u8,

    #[deku(
        ctx = "BufferKind::NullTerminated, 0",
        map = "EncodedStringBuffer::map"
    )]
    mime_type: String,

    picture_type: u8,

    #[deku(
        ctx = "BufferKind::NullTerminated, *encoding",
        map = "EncodedStringBuffer::map"
    )]
    description: String,

    #[deku(count = "size - (input.offset_from(rest) as u32 / 8)")]
    picture: Vec<u8>,
}
```

where `mime_type` and `description` are null-terminated strings and can therefore be any length, and the `picture` field is just "the rest of the binary data to fill the given field size". I am currently using `input.offset_from(rest)` to calculate this read size, however I do not know if this is actually _intended_ behaviour. Whether it is or not, I think I might be nice to expose this value as an available variable, so that this calculation can simply be something along the lines of `#[deku(count = "size - bytes_read")]`

sharksforarms:
Don't stop the issues! These are great ideas. I'm glad you were able to find a work-around.

This should include better documentation surrounding "internal" but available variables. Possibly finding a better way to expose/express this use case

samuelsleight:
Out of curiosity, is there an list somewhere of the other "internal" variables? It would be a good start to look at that if this is going to include better documentation for them all (which I agree it should)

sharksforarms:
> Out of curiosity, is there an list somewhere of the other "internal" variables?

What I do instead of tracing back proc-macro code is I directly look at the result after it's generated: `cargo expand --example example > examples/out.rs` and take a look at `from_bytes` inside of `out.rs`

What's available:
- `input` tuple
- Previous parsed fields
- `rest`
- (private vars via `__deku_`)

```rust
impl DekuContainerRead for DekuTest {
    fn from_bytes(input: (&[u8], usize)) -> Result<((&[u8], usize), Self), DekuError> {
        use core::convert::TryFrom;
        let input_bits = input.0.view_bits::<Msb0>();
        let mut rest = input.0.view_bits::<Msb0>();
        rest = &rest[input.1..];
        let __deku_field_a = {
[...]

PR #124, opened 2020-10-29
Implement attributes for the new container limit kinds

samuelsleight:
This PR implements the attributes for the new limit kinds introduced in the previous PR (#118), taking care of issues #76 and #110.

For #76, I'm conscious of the fact that a decision around the specific attrbute was never reached so I've arbitrarily opted for `bits_read` and `bytes_read` to avoid overloading any existing attributes, though I'm more than happy to change this if we decide something else is better.


deku-derive/src/macros/deku_read.rs
```diff
@@ -348,20 +348,31 @@ fn emit_field_read(
...
+        } else if let Some(field_bits) = &f.bits_read {
+            quote! {
+                {
+                    use core::borrow::Borrow;
+                    DekuRead::read(rest, (deku::ctx::BitSize(usize::try_from(*((#field_bits).borrow()))?).into(), (#read_args)))
+                }
+            }
+        } else if let Some(field_until) = &f.until {
+            // We wrap the input into another closure here to enforce that it is actually a callable
+            // Otherwise, an incorrectly passed-in integer could unexpectedly convert into a `Count` limit
```
sharksforarms, line 374: We could remove the `From/Into` impl on `Limit` and enforce an explicit `Limit` be initialized? This may make it more readable as well.

tests/test_attributes/test_vec_limits/test_bits_read.rs
```diff
@@ -0,0 +1,78 @@
...
+fn test_bits_read_from_field() {
+    #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
+    struct TestStruct {
+        bits: u8,
+
+        #[deku(endian = "little", bits_read = "bits")]
+        data: Vec<u16>,
+    }
+
+    let test_data: Vec<u8> = [0x10, 0xAA, 0xBB].to_vec();
```
sharksforarms, line 41: nit: for clarity: `0x10` -> `16`

tests/test_attributes/test_vec_limits/test_bits_read.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    let mut ret_read = TestStruct::try_from(test_data.as_ref()).unwrap();
+    assert_eq!(
+        TestStruct {
+            bits: 0x10,
```
sharksforarms, line 46: nit: for clarity: `0x10` -> `16`

tests/test_attributes/test_vec_limits/test_bits_read.rs
```diff
@@ -0,0 +1,78 @@
...
+        ret_read
+    );
+
+    // Add an item to the vec
+    ret_read.data.push(0xFFEE);
+
+    // `bits` is still 0x02, this is intended. `update` attribute should be
+    // used if `bits` is to be updated
+    let ret_write: Vec<u8> = ret_read.try_into().unwrap();
+    assert_eq!([0x10, 0xAA, 0xBB, 0xEE, 0xFF].to_vec(), ret_write);
```
sharksforarms, line 61: nit: for clarity: `0x10` -> `16`

tests/test_attributes/test_vec_limits/test_bits_read.rs
```diff
@@ -0,0 +1,78 @@
...
+fn test_bits_read_error() {
+    #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
+    struct TestStruct {
+        bits: u8,
+
+        #[deku(endian = "little", bits_read = "bits")]
+        data: Vec<u16>,
+    }
+
+    let test_data: Vec<u8> = [0x11, 0xAA, 0xBB].to_vec();
```
sharksforarms, line 75: nit: for clarity: `0x11` -> `17`

sharksforarms:
Small comments! Looks good to me!

Naming is hard.

deku-derive/src/macros/deku_read.rs
samuelsleight: The problem with that is that I _think_ it means we need to specify the `Limit` type parameters here, which is non-trivial to do?

I've just now thought that it might be nicer to make the attribute take an expression instead of a callable directly, and then this line becomes:
```rust
             quote! {DekuRead::read(rest, ((|value: &_| #field_until).into(), (#read_args)))} 
```

This feels much nicer, and also means the user doesn't need to explicity mark the type in their lambda, as they just have to specify an expression using `value`?

samuelsleight:
> Naming is hard.

Oh man, this is certainly true 😛  I really wish I could come up with something better for these bits/bytes attributes

deku-derive/src/macros/deku_read.rs
sharksforarms: Specifying a lambda makes it explicit from the user and no need to remember/know that `value` is available/exists. Comes back to the `internal available variables` discussion in https://github.com/sharksforarms/deku/issues/111

I prefer the long form in this case, if we can avoid having "internal" variables

sharksforarms: A possible alternative could be to write functions on the `Limit` enum and use them to enforce type arguments

```rust
impl Limit {
fn new_count(v: usize) {...}
fn new_bits(v: BitSize) {...}
fn new_until(v: FnMut) {...}
}
```

samuelsleight: > Specifying a lambda makes it explicit from the user and no need to remember/know that value is available/exists. 

Yeah, good point, it's much nicer to be explicit

> A possible alternative could be to write functions on the Limit enum and use them to enforce type arguments

For some reason I thought this wouldn't nicely work, but then I remembered it obviously does because `Vec::new` doesn't require explicit arguments to `Vec` 🤦

PR #135, opened 2020-11-19
Use intra-doc links

wcampbell0x2a:
With the release of 1.48.0, introduce the feature into this codebase.


sharksforarms:
This is great! Thanks!

PR #151, opened 2020-12-14
feat: Add `skip_bits` and `skip_bytes` attribute

### LINKED ISSUE #12: Add support for unused/padding bits
It would be nice if there was a way to skip over bits or bytes without creating dummy fields in order to save on space e.g : 
```Rust
pub struct SomeStruct {
    field_01: u8,

    // This field is useless but is needed for proper read/write
    unused01: u8,

    field_02: u8,

    // This field is useless but is needed for proper read/write
    unused02: u8,
}
```

maybe some `skip_[bytes|bits]` that could be added before and after fields in structs :
```Rust
#[deku(skip_bytes="1")]
```

sharksforarms:
Fixes #12

### DIFF
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -232,6 +232,12 @@ struct FieldData {
     /// skip field reading/writing
     skip: bool,
 
+    /// skip a number of bits
+    skip_bits: Option<TokenStream>,
+
+    /// skip a number of bytes
+    skip_bytes: Option<TokenStream>,
+
     /// read field as temporary value, isn't stored
     temp: bool,
 
@@ -271,6 +277,8 @@ impl FieldData {
             reader: receiver.reader,
             writer: receiver.writer,
             skip: receiver.skip,
+            skip_bits: receiver.skip_bits,
+            skip_bytes: receiver.skip_bytes,
             temp: receiver.temp,
             default,
             cond: receiver.cond,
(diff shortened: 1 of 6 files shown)


sharksforarms:
Still unsure about this one. Currently this only skips on read, should it "skip" or "pad" with something on write? Thoughts? @passcod

inspier:
Personally, having it also "pad" on write would be nice, since generally for the formats I am working with, you want the padding to be preserved on serialization though the actual contents of the padding is unimportant. Strictly speaking you can implement this pad on write with custom writers too so thats not really a problem.

passcod:
Certainly pad on write by default (with a custom writer to opt out), otherwise I think it's much less intuitive

sharksforarms:
@inspier @passcod 

Thanks for the feedback. In that case, I propose 4 attributes instead:
- pad_bits_before: pad x bits before reading/writing (kinda long name...?)
- pad_before: pad x bytes before reading/writing
- pad_bits_after: pad x after reading/writing
- pad_after: pad x bytes after reading/writing
- pad_with: default: 0x00 for bytes, false for bits

passcod:
That's comprehensive and covers everything I can think of. Only thing I'd say is I'm not sure if it should be `pad_after` or `pad_bytes_after`... the latter keeps in better with the other attributes that have bits and bytes versions without a clear "default"

sharksforarms:
> That's comprehensive and covers everything I can think of. Only thing I'd say is I'm not sure if it should be `pad_after` or `pad_bytes_after`... the latter keeps in better with the other attributes that have bits and bytes versions without a clear "default"

Yeahh that's what I had it at first, was worried about the explicitness where most will be dealing with bytes over bits

passcod:
Hmm, well, I can only offer my experience, where most of the cases I wish for it are when I read 2 bits and skip 6, that kinda thing. However, I don't have strong feelings either way.

sharksforarms:
Closing in favour of the above attributes, stay tuned!

Issue #141, opened 2020-11-30
Obtain id of enum variant at "runtime"


passcod:
It would be nice to be able to obtain the deku `id` of a given enum variant from a method on the enum, for testing and other purposes (custom writers?). E.g.

```rust
#[derive(DekuRead)]
#[deku(type = "u8")]
enum Request {
	#[id = "0x01"]
	Cats { toy: Toy },

	#[id = "0x10"]
	Dogs { ball: Ball },
}

fn test_request() {
	let request = vec![Request::Cats { toy }.deku_id()];
	Request::from_bytes((&request, 0))?; // ...
}
```

Currently I make do with duplicating the IDs in a match in a method.

wcampbell0x2a:
You may want to look at: https://github.com/sharksforarms/deku/pull/139

passcod:
That won't work with non-unit enums, will it?

wcampbell0x2a:
> That won't work with non-unit enums, will it?

True

wcampbell0x2a:
@passcod Current there is a MR(https://github.com/sharksforarms/deku/pull/176) for this, if you have any input lmk

sharksforarms:
Now available in 0.11.0

PR #176, opened 2021-02-17
Add deku_id()

wcampbell0x2a:
This creates a `deku_id()` function for enums that gives a user the ability for knowing the id of a enum variant at runtime. See https://github.com/sharksforarms/deku/issues/141

Any comments you have would be great! I'm still new to more advanced TokenStream _fun_.

- Having some issues with implementing the Generic part of the enum, and not sure how to implement that part of the code generation.
```
error[E0412]: cannot find type `T` in this scope
  --> src/lib.rs:89:15
   |
89 | enum TestEnum<T>
   |               ^ not found in this scope

error: aborting due to previous error
```
- Not sure on if you want this to be a trait function grouped with `DekuRead`?
- I have more testing I can throw in that enum test file

### DIFF
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -128,7 +128,6 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
         });
     }
 
-    // println!("{}", tokens.to_string());
     Ok(tokens)
 }
 
@@ -164,6 +163,7 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
     let mut has_default_match = false;
     let mut pre_match_tokens = vec![];
     let mut variant_matches = vec![];
+    let mut deku_ids = vec![];
 
     let has_discriminant = variants.iter().any(|v| v.discriminant.is_some());
 
@@ -219,6 +219,16 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
             let internal_fields = gen_internal_field_idents(variant_is_named, field_idents);
             let initialize_enum =
                 super::gen_enum_init(variant_is_named, variant_ident, internal_fields);
+
+            if !has_discriminant && variant.id_pat.is_none() {
+                let variant_deku_id = variant.id.as_ref();
+                let deku_id = if let Id::LitByteStr(variant_deku_id) = variant_deku_id.unwrap() {
+                    quote! { Self :: #initialize_enum => *#variant_deku_id}
+                } else {
+                    quote! { Self :: #initialize_enum => #variant_deku_id}
+                };
+                deku_ids.push(deku_id);
+            }
 
             // if we're consuming an id, set the rest to new_rest before reading the variant
             let new_rest = if consume_id {
@@ -352,7 +362,36 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
         });
     }
 
-    // println!("{}", tokens.to_string());
+    // create `deku_id` function for retrieving at runtime the id value
+    let deku_id_id_type = if id_type.is_some() || input.ctx.is_none() {
+        id_type.unwrap()
+    } else {
+        &ctx_types
+    };
+    let deku_id = if deku_ids.is_empty() {
+        quote! {
+            impl #ident {
+                fn deku_id(&self) -> #deku_id_id_type {
+                    match self {
+                        _ => panic!("deku id for variant not found")
+                    }
+                }
+            }
+        }
+    } else {
+        quote! {
+            impl #ident {
+                fn deku_id(&self) -> #deku_id_id_type {
+                    match self {
+                        #(#deku_ids),*,
+                        _ => panic!("deku id for variant not found"),
+                    }
+                }
+            }
+        }
+    };
+    tokens.extend(deku_id);
+
     Ok(tokens)
 }
 
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -371,7 +371,7 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
         &ctx_types
     };
     let deku_id = quote! {
-        impl #ident {
+        impl #imp DekuEnumExt<#lifetime, #deku_id_id_type> for #ident #wher {
             fn deku_id(&self) -> Result<#deku_id_id_type, DekuError> {
                 match self {
                     #(#deku_ids ,)*
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -224,9 +224,9 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
             if !has_discriminant && variant.id_pat.is_none() {
                 let variant_deku_id = variant.id.as_ref();
                 let deku_id = if let Id::LitByteStr(variant_deku_id) = variant_deku_id.unwrap() {
-                    quote! { Self :: #initialize_enum => *#variant_deku_id}
+                    quote! { Self :: #initialize_enum => Ok(*#variant_deku_id)}
                 } else {
-                    quote! { Self :: #initialize_enum => #variant_deku_id}
+                    quote! { Self :: #initialize_enum => Ok(#variant_deku_id)}
                 };
                 deku_ids.push(deku_id);
             }
@@ -363,7 +363,7 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
         });
     }
 
-    // create `deku_id` function for retrieving at runtime the id value
+    // create `deku_id()` function for retrieving at runtime the id value
     let deku_id_id_type = if let Some(id_type) = id_type {
         id_type
     } else {
@@ -372,9 +372,9 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
     let deku_id = if deku_ids.is_empty() {
         quote! {
             impl #ident {
-                fn deku_id(&self) -> #deku_id_id_type {
+                fn deku_id(&self) -> Result<#deku_id_id_type, DekuError> {
                     match self {
-                        _ => panic!("deku id for variant not found")
+                        _ => Err(DekuError::IdVariantNotFound),
                     }
                 }
             }
@@ -382,10 +382,10 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
     } else {
         quote! {
             impl #ident {
-                fn deku_id(&self) -> #deku_id_id_type {
+                fn deku_id(&self) -> Result<#deku_id_id_type, DekuError> {
                     match self {
                         #(#deku_ids),*,
-                        _ => panic!("deku id for variant not found"),
+                        _ => Err(DekuError::IdVariantNotFound),
                     }
                 }
             }
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -1,7 +1,8 @@
 use crate::{
     macros::{
         gen_ctx_types_and_arg, gen_field_args, gen_id_args, gen_internal_field_ident,
-        gen_internal_field_idents, pad_bits, gen_type_from_ctx_id, token_contains_string, wrap_default_ctx,
+        gen_internal_field_idents, gen_type_from_ctx_id, pad_bits, token_contains_string,
+        wrap_default_ctx,
     },
     Id,
 };
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -370,24 +370,12 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
     } else {
         &ctx_types
     };
-    let deku_id = if deku_ids.is_empty() {
-        quote! {
-            impl #ident {
-                fn deku_id(&self) -> Result<#deku_id_id_type, DekuError> {
-                    match self {
-                        _ => Err(DekuError::IdVariantNotFound),
-                    }
-                }
-            }
-        }
-    } else {
-        quote! {
-            impl #ident {
-                fn deku_id(&self) -> Result<#deku_id_id_type, DekuError> {
-                    match self {
-                        #(#deku_ids),*,
-                        _ => Err(DekuError::IdVariantNotFound),
-                    }
+    let deku_id = quote! {
+        impl #ident {
+            fn deku_id(&self) -> Result<#deku_id_id_type, DekuError> {
+                match self {
+                    #(#deku_ids ,)*
+                    _ => Err(DekuError::IdVariantNotFound),
                 }
             }
         }
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -364,8 +364,8 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
     }
 
     // create `deku_id` function for retrieving at runtime the id value
-    let deku_id_id_type = if id_type.is_some() || input.ctx.is_none() {
-        id_type.unwrap()
+    let deku_id_id_type = if let Some(id_type) = id_type {
+        id_type
     } else {
         &ctx_types
     };
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -128,6 +128,7 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
         });
     }
 
+    // println!("{}", tokens.to_string());
     Ok(tokens)
 }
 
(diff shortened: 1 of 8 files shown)


deku-derive/src/macros/deku_read.rs
```diff
@@ -128,7 +128,6 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
         });
     }
 
-    // println!("{}", tokens.to_string());
```
sharksforarms, line 131: The comment print can stay :) it's useful for debugging the output when cargo expand fails. I can add a comment to it.

deku-derive/src/macros/deku_read.rs
```diff
@@ -220,6 +220,16 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
             let initialize_enum =
                 super::gen_enum_init(variant_is_named, variant_ident, internal_fields);
 
+            if !has_discriminant && variant.id_pat.is_none() {
```
sharksforarms, line 223: It might be possible to do this section earlier when we're figuring out the variant id:
https://github.com/sharksforarms/deku/blob/afa7672dbf86977a93eda5ed874d0bc77aa4171d/deku-derive/src/macros/deku_read.rs#L179-L183

deku-derive/src/macros/deku_read.rs
```diff
@@ -352,7 +362,36 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
         });
     }
 
-    // println!("{}", tokens.to_string());
```
sharksforarms, line 355: (print can stay as per last comment)

deku-derive/src/macros/deku_read.rs
```diff
@@ -352,7 +362,36 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
...
+    // create `deku_id` function for retrieving at runtime the id value
+    let deku_id_id_type = if id_type.is_some() || input.ctx.is_none() {
+        id_type.unwrap()
```
sharksforarms, line 367: this could panic if `id_type==None` and `input.ctx==Some`

Would perfer an `if let` binding

deku-derive/src/macros/deku_read.rs
```diff
@@ -352,7 +362,36 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
...
+    } else {
+        &ctx_types
```
sharksforarms, line 369: is there a test case for this case?

deku-derive/src/macros/deku_read.rs
```diff
@@ -352,7 +362,36 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
...
+    };
+    let deku_id = if deku_ids.is_empty() {
+        quote! {
+            impl #ident {
+                fn deku_id(&self) -> #deku_id_id_type {
+                    match self {
+                        _ => panic!("deku id for variant not found")
```
sharksforarms, line 376: We should avoid panics, a `DekuError` should be returned

deku-derive/src/macros/deku_read.rs
```diff
@@ -352,7 +362,36 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
...
+                    }
+                }
+            }
+        }
+    } else {
+        quote! {
+            impl #ident {
+                fn deku_id(&self) -> #deku_id_id_type {
+                    match self {
+                        #(#deku_ids),*,
```
sharksforarms, line 386: Since this is the only difference in the if case, I think we can collapse them into one
If `deku_ids` is empty, this should not even expand, leaving us with the first case :)

```suggestion
                        #(#deku_ids),*
```

tests/test_enum.rs
```diff
@@ -86,16 +86,17 @@ fn test_enum_discriminant(input: &[u8], expected: TestEnumDiscriminant) {
...
+enum TestEnumArray {
+    #[deku(id = b"123")]
+    VarA,
+    #[deku(id = "[1,1,1]")]
+    VarB,
+}
+
 #[test]
 fn test_enum_array_type() {
-    #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
```
sharksforarms, line 91: I would prefer to see another struct inside another test function instead of re-using this

sharksforarms (changes requested):
Thanks @wcampbell0x2a ! Just some general comments on how to make it better. This looks great!

(Also PR is coming to fix bitvec/clippy/build errors... sorry about that)

Edit: Done. You can now rebase with master and the builds should be better

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: Well, with this diff:
```
diff --git a/deku-derive/src/macros/deku_read.rs b/deku-derive/src/macros/deku_read.rs
index b587736..54ffaf0 100644
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -368,30 +368,19 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
     } else {
         &ctx_types
     };
-    let deku_id = if deku_ids.is_empty() {
-        quote! {
-            impl #ident {
-                fn deku_id(&self) -> #deku_id_id_type {
-                    match self {
-                        _ => panic!("deku id for variant not found")
-                    }
-                }
-            }
-        }
-    } else {
-        quote! {
-            impl #ident {
-                fn deku_id(&self) -> #deku_id_id_type {
-                    match self {
-                        #(#deku_ids),*,
-                        _ => panic!("deku id for variant not found"),
-                    }
+    let deku_id = quote! {
+        impl #ident {
+            fn deku_id(&self) -> #deku_id_id_type {
+                match self {
+                    #(#deku_ids),*
+                    _ => panic!("deku id for variant not found"),
                 }
             }
         }
     };
     tokens.extend(deku_id);
```
     
It looks like the last item doesn't print the trailing `,`
```
} impl TestEnumArray
{
    fn deku_id(& self) -> [u8 ; 3]
    {
        match self
        {
            Self :: VarA => * b"123", Self :: VarB => [1, 1, 1] _ => panic !
            ("deku id for variant not found"),
        }
    }
}
```

sharksforarms: Try this: `#(#var ,)*`

https://github.com/dtolnay/quote#repetition

tests/test_enum.rs
wcampbell0x2a: Refactored all tests for this MR into `tests/test_deku_id.rs`

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: Fixed

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: Should be tested in `tests/test_deku_id.rs`

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: Fixed

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: Added Option return for the let assignment @ 179, I think it's much nicer.

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: sweet, works. thanks.

wcampbell0x2a:
Only thing left is the `tests/test_generic.rs`, any hints?

wcampbell0x2a:
Also adding docs, do you have an area in which you think docs for this function would be applicable?

passcod:
This make `deku_id()` an inherent method, right? Wouldn't it be better to have on a dedicated trait instead so it doesn't conflict with any user-defined `deku_id()`?

deku-derive/src/macros/deku_read.rs
```diff
@@ -352,6 +364,24 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
...
     }
 
+    // create `deku_id()` function for retrieving at runtime the id value
+    let deku_id_id_type = if let Some(id_type) = id_type {
+        id_type
+    } else {
+        &ctx_types
+    };
+    let deku_id = quote! {
+        impl #ident {
```
sharksforarms, line 374: I think a `DekuEnumExt` trait to hold these functions would be great as @passcod mentioned

src/error.rs
```diff
@@ -41,6 +41,8 @@ pub enum DekuError {
     Unexpected(String),
     /// Assertion error from `assert` or `assert_eq` attributes
     Assertion(String),
+    /// `id` attribute not found in `deku_id()`
```
sharksforarms, line 44: "Could not resolve `id` for variant"

src/error.rs
```diff
@@ -74,6 +76,7 @@ impl core::fmt::Display for DekuError {
             DekuError::InvalidParam(ref err) => write!(f, "Invalid param error: {}", err),
             DekuError::Unexpected(ref err) => write!(f, "Unexpected error: {}", err),
             DekuError::Assertion(ref err) => write!(f, "Assertion error: {}", err),
+            DekuError::IdVariantNotFound => write!(f, "deku id for variant not found"),
```
sharksforarms, line 79: "Could not resolve `id` for variant"

tests/test_deku_id.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    assert_eq!(Ok(Request2::Cats), Request3::Cats.deku_id());
+    assert_eq!(Ok(Request2::Dogs), Request3::Dogs.deku_id());
+}
+
+#[test]
+fn test_ctx() {
+    #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
+    struct StructEnumId {
```
sharksforarms, line 48: not used

tests/test_deku_id.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    #[derive(Copy, Clone, PartialEq, Debug, DekuRead, DekuWrite)]
+    #[deku(type = "u8")]
+    enum Nice {
+        True = 0x00,
+        False = 0x01,
+    }
+
+    #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
+    struct StructEnumId2 {
```
sharksforarms, line 74: not used

tests/test_deku_id.rs
```diff
@@ -0,0 +1,118 @@
...
+    enum TopLevelCtxEnum {
+        #[deku(id = "1")]
+        VariantA(u8),
+    }
+
+    assert_eq!(Ok(1), TopLevelCtxEnum::VariantA(0).deku_id());
+}
+
+#[test]
+fn test_advanced() {
```
sharksforarms, line 107: `test_litbytestr`

deku-derive/src/macros/deku_read.rs
sharksforarms: Agreed!

sharksforarms:
> Only thing left is the `tests/test_generic.rs`, any hints?

Here is an example:

https://github.com/sharksforarms/deku/blob/master/deku-derive/src/macros/deku_read.rs#L658

> Also adding docs, do you have an area in which you think docs for this function would be applicable?

Let's see what the docs look like with a `DekuEnumExt` trait, this may be sufficient. We can also expand the `example/enums.rs` to include it.

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: weird that tarpaulin says this isn't tested, but it is 3 times :shrug:

wcampbell0x2a:
Fixed generics, added to `examples/enums.rs` and added trait :rocket:

deku-derive/src/macros/deku_read.rs
```diff
@@ -220,6 +225,13 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
             let initialize_enum =
                 super::gen_enum_init(variant_is_named, variant_ident, internal_fields);
 
+            if let Some(deku_id_deref) = deku_id_deref {
+                let variant_deku_id = variant.id.as_ref();
+                let deku_id =
+                    quote! { Self :: #initialize_enum => Ok(#deku_id_deref #variant_deku_id)};
+                deku_ids.push(deku_id);
+            }
+
```
sharksforarms, line 234: It may make more sense to do the deref logic here:

```suggestion
            if let Some(variant_id) = &variant.id {
                let deref = match variant_id {
                    Id::TokenStream(v) => quote!{},
                    Id::LitByteStr(v) => quote! {*},
                };
                let deku_id =
                    quote! { Self :: #initialize_enum => Ok(#deref #variant_id)};
                deku_ids.push(deku_id);
            }

```

deku-derive/src/macros/deku_read.rs
```diff
@@ -352,6 +364,24 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
         });
     }
 
+    // create `deku_id()` function for retrieving at runtime the id value
```
sharksforarms, line 367: update comment

src/lib.rs
```diff
@@ -330,3 +330,9 @@ pub trait DekuUpdate {
     /// Apply updates
     fn update(&mut self) -> Result<(), DekuError>;
 }
+
+/// "Extended Enum" trait: obtain additional enum information
+pub trait DekuEnumExt<'a, T> {
```
sharksforarms, line 335: Docs look good with this :) nice!

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: Fixed

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: Fixed

deku-derive/src/macros/deku_read.rs
```diff
@@ -352,6 +364,24 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
...
+    let deku_id_id_type = if let Some(id_type) = id_type {
+        id_type
+    } else {
+        &ctx_types
```
sharksforarms, line 370: The problem with using `ctx_types` is when there's more than one, it breaks:

```rust
#[deku(ctx = "test: u16, my_id: u8", id = "my_id")]
enum EnumId {
    #[deku(id = "1")]
    VarA(u8),
    #[deku(id = "2")]
    VarB,
}
```

generates

```rust
impl DekuEnumExt<'_, (u16, u8)> for EnumId {
    fn deku_id(&self) -> Result<(u16, u8), DekuError> {
        match self {
            Self::VarA(__deku_field_0) => Ok(1),
            Self::VarB => Ok(2),
            _ => Err(DekuError::IdVariantNotFound),
        }
    }
}
```

One way of solving this would be to find the type of `input.id` in `input.ctx`

#[deku(ctx = "test: u16, **my_id**: **u8**", id = "**my_id**")]

Having a utility function in `macros/mod.rs` would be useful. You can look at `gen_ctx_types_and_arg` to see how it iterates `ctx`

tests/test_deku_id.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    #[derive(Copy, Clone, PartialEq, Debug, DekuRead, DekuWrite)]
+    #[deku(type = "u8")]
+    enum Nice {
+        True = 0x00,
+        False = 0x01,
+    }
+
+    #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
+    #[deku(ctx = "my_id: Nice", id = "my_id")]
```
sharksforarms, line 66: This seems to be testing the same thing as the previous, possibly change to include 2 ctx arguments

tests/test_enum.rs
```diff
@@ -85,23 +85,3 @@ fn test_enum_discriminant(input: &[u8], expected: TestEnumDiscriminant) {
     let ret_write: Vec<u8> = ret_read.try_into().unwrap();
     assert_eq!(input.to_vec(), ret_write);
 }
-
-#[test]
```
sharksforarms, line 89: This test can stay

tests/test_deku_id.rs
```diff
@@ -100,3 +100,15 @@ fn test_litbytestr() {
 
     assert_eq!(b"123", TestEnumArray::VarA.deku_id().unwrap().as_ref());
 }
+
+#[test]
+#[should_panic(expected = "called `Result::unwrap()` on an `Err` value: IdVariantNotFound")]
```
sharksforarms, line 105: Good test, I'm sure this could be supported with deku_id() but I'm not sure it's really needed as the discriminant is already available? If anything, it would be done for consistency... Could be done in a separate PR.

wcampbell0x2a: That was my thoughts exactly on that issue.

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: Just pushed https://github.com/sharksforarms/deku/pull/176/commits/649b1d323621fcae5367208fb78c434cee3d0dd0 with this fix. It works and the tests compile but the code quality might be lacking, any opinions or extra tests are good ;)

sharksforarms (approved):
Looks great! Thanks @wcampbell0x2a

PR #187, opened 2021-03-15
add optional `const_generics` support

soruh:
Adds a `const_generics` feature that uses `min_const_generics` which will be stabilized on March 25, 2021 as part of Rust 1.51 (and is currently available in beta) to allow for reading/writing arrays with >32 elements when enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ members = [
 default = ["std"]
 std = ["deku_derive/std", "bitvec/std", "alloc"]
 alloc = ["bitvec/alloc"]
+const_generics = []
 
 [dependencies]
 deku_derive = { version = "^0.11.0", path = "deku-derive", default-features = false}
--- a/src/impls/slice.rs
+++ b/src/impls/slice.rs
@@ -89,74 +89,140 @@ where
     }
 }
 
-macro_rules! ImplDekuSliceTraits {
-    ($typ:ty; $($count:expr),+ $(,)?) => {
+#[cfg(not(feature = "const_generics"))]
+mod pre_const_generics_impl {
+    use super::*;
 
-        impl<Ctx: Copy> DekuWrite<Ctx> for &[$typ]
-        where
-            $typ: DekuWrite<Ctx>,
-        {
-            fn write(&self, output: &mut BitVec<Msb0, u8>, ctx: Ctx) -> Result<(), DekuError> {
-                for v in *self {
-                    v.write(output, ctx)?;
-                }
-                Ok(())
-            }
-        }
+    macro_rules! ImplDekuSliceTraits {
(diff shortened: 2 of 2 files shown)


sharksforarms (approved):
This looks great! Thanks!

soruh:
great, thank you!

v1gnesh:
Hey @soruh, could you show an example of how this can be used?

wcampbell0x2a:
> Hey @soruh, could you show an example of how this can be used?

This feature is enabled on master: https://github.com/sharksforarms/deku/blob/d04559605bed4476b1a78aaffd843377d6ab8fe6/Cargo.toml#L22

soruh:
It allows you to use arrays that are more than 32 elements long and arrays with const generic sizes, so [this impl](https://docs.rs/deku/latest/deku/trait.DekuRead.html#impl-DekuRead%3C'a,+Ctx%3E-for-%5BT;+N%5D)

here's an example I guess:
```rust
use deku::{DekuContainerRead, DekuContainerWrite, DekuRead, DekuUpdate, DekuWrite};

#[deku(endian = "big")]
struct DekuTest<const N: usize> {
    #[deku(bits = "4")]
    field_a: u8,
    #[deku(bits = "4")]
    field_b: u8,
    field_c: [u8; 40],
    field_d: [u8; N],
}

fn main() {
    const N: usize = 128;

    let mut data = vec![0u8; 1 + 40 + N];
    data[0] = 170;
    data[1..]
        .iter_mut()
        .enumerate()
        .for_each(|(i, x)| *x = i as u8);
    let (_, value) = DekuTest::<N>::from_bytes((data.as_ref(), 0)).unwrap();

    dbg!(value);
}
```

v1gnesh:
> > Hey @soruh, could you show an example of how this can be used?
> 
> This feature is enabled on master:
> 
> https://github.com/sharksforarms/deku/blob/d04559605bed4476b1a78aaffd843377d6ab8fe6/Cargo.toml#L22

Do I have to add something to my project's Cargo.toml to make use of this?
I get a squiggly red line under DekuRead when my Struct uses has a field like this:

```rust
    field: MoString<n>
```

where

```rust
pub struct MoString<const N: usize> (
    #[deku(map=func)
    String,
);
```

soruh:
The feature should is enabled by default, unless you explicitly turned it of with default-features=false or are using an ancient version. I think you probably made a mistake while implementing your MyString type. Also generally, a merged pull request from two years ago is not the greatest place to ask these things... Maybe open a [discussion](https://github.com/sharksforarms/deku/discussions/categories/general) or something?

v1gnesh:
Yes, sure, will try to reproduce separately and open a discussion. Thanks for responding here.

PR #189, opened 2021-03-25
deps: Bump hexlit to 0.5.0 and fix broken test on latest stable.

inspier:
This PR bumps hexlit to 0.5.0 which offers some small speed and size improvements due to min_const_generics now being stablized. It also fixes a test that used a temporary value which could be dropped while borrowed which is a hard error on the latest stable (1.51).


sharksforarms:
looks like the benchmarks use 1.50.0 still hm

sharksforarms:
Do you mind rebasing on latest master to test benchmark changes? Should use latest stable now

inspier:
Sure thing! Sorry for the trouble.

sharksforarms (approved):

PR #196, opened 2021-04-06
Add docs on the return type for DekuRead

caass:
When trying to write my own implementation of DekuRead, instead of `#[derive]`ing it,
I was a little confused about the function signature of `DekuRead::read`.
Adding the return type to the documentation makes it easier for folks to write
their own implementations in cases where the provided tooling
(which is very good and extensive, btw) doesn't suffice.

FWIW my specific use-case involves a bitfield that's of variable length, and I couldn't think
of a simple way to use the `#[deku]` attributes to make it work in an elegant way.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -285,7 +285,7 @@ pub use crate::error::DekuError;
 
 /// "Reader" trait: read bits and construct type
 pub trait DekuRead<'a, Ctx = ()> {
-    /// Read bits and construct type
+    /// Read bits and construct type. Returns the remaining bits after parsing in addition to Self.
     /// * **input** - Input as bits
     /// * **ctx** - A context required by context-sensitive reading. A unit type `()` means no context
     /// needed.


caass:
Also, it's probably worth adding an example of a custom implementation of `DekuRead` and `DekuWrite` to the docs

sharksforarms:
Thanks! I made some other modifications here: https://github.com/sharksforarms/deku/commit/fcfdc24eca1b8663f9a2cd2d0d8ad6534b08a862

sharksforarms:
I created this ticket for the DekuRead/DekuWrite implementation: https://github.com/sharksforarms/deku/issues/197

Out of curiosity, is the data format you're parsing shareable? I'd like to understand your use-case better. Pseudo code/format works too.

caass:
Yeah sure! I'm trying to write a parser for [RTMP](https://wwwimages2.adobe.com/content/dam/acom/en/devnet/rtmp/pdf/rtmp_specification_1.0.pdf). The part I wound up implementing myself was the [Chunk Basic Header](https://wwwimages2.adobe.com/content/dam/acom/en/devnet/rtmp/pdf/rtmp_specification_1.0.pdf#page=12).

Essentially the issue is that the header is either one, two, or three bytes. And the length of the header is determined by the last six bits of the first byte. So I could either define a bunch of `Option`s for the one-, two-, or three-byte versions, or write my own implementation.

[Here's](https://gist.github.com/caass/1467d723a8dbd98152e808b00b482735) the implementation I wound up writing. Feel free to use this as your example for the docs, although it may be worth generify-ing it slightly so you don't need any background on the RTMP Chunk Stream to understand what's going on.

Also -- this library is awesome! I tried writing an RTMP parser by hand some months ago, and Deku saves me so much time, and it's always correct. Huge props ❤️

caass:
@sharksforarms pinging just in case u didn't see this 🙈 forgive me

sharksforarms:
Thanks @caass for the kind words I appreciate it! I've added your example to the issue :+1:

PR #198, opened 2021-04-06
feat: Impl for tuple

abungay:
Implements DekuRead + DekuWrite for tuples of length 1 to 11.

### DIFF
--- a/src/impls/tuple.rs
+++ b/src/impls/tuple.rs
@@ -1,0 +1,119 @@
+//! Implementations of DekuRead and DekuWrite for tuples of length 1 to 11
+
+use crate::{DekuError, DekuRead, DekuWrite};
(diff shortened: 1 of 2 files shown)


src/impls/tuple.rs
```diff
@@ -0,0 +1,120 @@
+//! Implementations of DekuRead and DekuWrite for tuples of length 1 to 11
+
+use crate::{DekuError, DekuRead, DekuWrite};
+use bitvec::prelude::*;
+pub use deku_derive::*;
```
sharksforarms, line 5: I think this line can be removed

abungay: removed line :)

sharksforarms:
Thanks!

PR #200, opened 2021-04-10
Fix CI failure for Codecov/tarpaulin

abungay:
Run tarpaulin for code coverage on stable rust instead of nightly.

Nightly rust has our test cases failing due to the changed debug
printing of strings with single quotes "'" so our test cases using
should_panic(expected = <error message>) are failing.


sharksforarms:
Thanks! :tada:

PR #203, opened 2021-04-14
Use AsRef in `DekuContainerRead::from_bytes`

### LINKED ISSUE #48: Use AsRef instead of &[u8]

constfold:
Fix #48. Tests not updated. All failed tests are caused by the manual use of `as_ref`. e.g. https://github.com/sharksforarms/deku/blob/e46d9502593417cb92d18f0bc130dff1180abe50/src/lib.rs#L176
Note: This change will break all codes that calls `as_ref` in argument passing because of the type inference issue (can't resolve two generics in `AsRef` at the same time).

### DIFF
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -688,7 +688,8 @@ pub fn emit_from_bytes(
     quote! {
         impl #imp ::#crate_::DekuContainerRead<#lifetime> for #ident #wher {
             #[allow(non_snake_case)]
-            fn from_bytes(__deku_input: (&#lifetime [u8], usize)) -> Result<((&#lifetime [u8], usize), Self), ::#crate_::DekuError> {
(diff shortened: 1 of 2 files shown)


sharksforarms:
Hey @constfold ! Glad to see your name pop up again. I am okay with the breaking change (pre-1.0) as long as it adds value which I believe this does.

constfold:
This change seems like no sense, since the signature need to be `(&'a (impl AsRef<[u8]> + ?Sized)` that can't receive a owned type which is the original purpose of the use `impl AsRef`. And `impl AsRef<[u8]> + 'lifetime` will failed to compile when the lifetime in `DekuContainerRead` is anonymous. 
Is it possible to make all `lifetime`s passed to `emit_from_bytes` has a name?

constfold:
Finally i found using `AsRef<[u8]>` is impossible or meaningless. So closed.

Issue #209, opened 2021-05-11
Test on big endian


sharksforarms:
Add testing on big endian to CI

Using miri?
https://github.com/rust-lang/miri#cross-interpretation-running-for-different-targets

PR #211, opened 2021-05-11
Add Miri tests including big endian

sharksforarms:
https://github.com/sharksforarms/deku/issues/209

Tests fail on big endian :(


sharksforarms:
May just be the way the tests are written? `TestPrimitive!(test_u16, u16, vec![0xABu8, 0xCD], 0xCDAB /* < expected value depends on endianess */);`

PR #226, opened 2021-06-13
Initial version of `as` attribute

### LINKED ISSUE #225: `#[deku(as = "")]`/`#[deku(with = "")]` attributes
Sort-of related to #174, but more general.

I'm looking into converting RustPython's bytecode format to use deku over bincode (https://github.com/RustPython/RustPython/issues/2362#issuecomment-857747216), and I'm running into a few issues cause I want to keep the `DekuRead/Write` structs roughly as they are, i.e. ergonomic to use from Rust. For example, not adding separate length fields for deserializing vectors, and also I'd like to use LEB128 to represent most integers as our existing bincode impl does. I understand why those aren't defaults in `deku`, but it'd be a lot easier to customize `deku` to meet our needs if there was something like serde's `with` attribute, or even `serde_with`'s [`as` attribute](https://docs.rs/serde_with/1.9.2/serde_with/guide/index.html#1-a-more-flexible-and-composable-replacement-for-the-with-annotation-called-serde_as-v150). Would you accept a PR that adds `Deku{Read/Write}As` traits and an `as` attribute to the derive macros so that individual field [de]serializing can be customized?

coolreader18:
Resolves #225.

Here's an initial draft of the "as" attribute, let me know what you think. Note that as of tomorrow, I'll be going to summer camp, so I may not be able to work on this very much; sorry about that :sweat_smile: It should really just be docs though, unless you want this initial PR to contain types that implement `DekuRead/WriteAs` as well.

### DIFF
--- a/tests/test_as.rs
+++ b/tests/test_as.rs
@@ -1,0 +1,160 @@
+use deku::ctx::Size;
+use deku::prelude::*;
+use deku::{bitvec, DekuReadAs, DekuWriteAs, Same};
+use std::marker::PhantomData;
+
+#[derive(DekuRead, DekuWrite, PartialEq, Debug)]
+struct Foo {
+    #[deku(as = "VecWithLen<Same>", bytes = "4", ctx = "()")]
+    ints: Box<[u32]>,
+}
+
+#[test]
+fn test_roundtrip() {
+    let foo = Foo {
+        ints: Box::new([6, 1, 25, 9]),
+    };
+    let bytes = foo.to_bytes().unwrap();
+    let foo2 = Foo::from_bytes((&bytes, 0)).unwrap().1;
+    assert_eq!(foo2, foo);
+}
+
+#[test]
+fn test_vecwithlen() {
+    let foo = Foo {
+        ints: Box::new([6, 1, 25, 9]),
+    };
+    let bytes = foo.to_bytes().unwrap();
+    #[rustfmt::skip]
+    assert_eq!(
+        bytes,
+        [
+            4,  0, 0, 0,
+            6,  0, 0, 0,
+            1,  0, 0, 0,
+            25, 0, 0, 0,
+            9,  0, 0, 0
+        ]
+    );
+}
+
+struct VecWithLen<T> {
+    _marker: PhantomData<T>,
+}
+
+impl<'a, Ctx: Copy, T, U: DekuReadAs<'a, T, Ctx>> DekuReadAs<'a, Box<[T]>, (Size, Ctx)>
+    for VecWithLen<U>
+{
+    fn read_as(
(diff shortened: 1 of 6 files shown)


sharksforarms:
This is neat! Thanks for making a PR. I'll give it a review shortly!

I propose adding an example, here's something I came up with to try it out. Feel free to add it to the PR.

```rust
use deku::prelude::*;
use std::net::Ipv4Addr;

struct Data {
    // Type does not implement DekuRead or DekuWrite
    #[deku(as = "MyIpAddr")]
    address: Ipv4Addr,
}

impl<'a> deku::DekuReadAs<'a, Ipv4Addr> for MyIpAddr {
    fn read_as(
        input: &'a deku::bitvec::BitSlice<deku::bitvec::Msb0, u8>,
        ctx: (),
    ) -> Result<(&'a deku::bitvec::BitSlice<deku::bitvec::Msb0, u8>, Ipv4Addr), DekuError> {
        u32::read(input, ctx).map(|(rest, v)| (rest, Ipv4Addr::from(v)))
    }
}

impl deku::DekuWriteAs<Ipv4Addr> for MyIpAddr {
    fn write_as(
        source: &Ipv4Addr,
        output: &mut deku::bitvec::BitVec<deku::bitvec::Msb0, u8>,
        ctx: (),
    ) -> Result<(), DekuError> {
        let ip: u32 = (*source).into();
        ip.write(output, ctx)
    }
}

struct MyIpAddr {}

fn main() {
    let data = Data {
        address: "127.0.0.1".parse().unwrap(),
    };

    let data_write = data.to_bytes().unwrap();
    assert_eq!(vec![1, 0, 0, 127], data_write);

    let (_rest, data_read) = Data::from_bytes((&data_write, 0)).unwrap();
    assert_eq!(data, data_read);
}
```

tests/test_as.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+impl<Ctx: Copy, T, U: DekuWriteAs<T, Ctx>> DekuWriteAs<[T], (Size, Ctx)> for VecWithLen<U> {
+    fn write_as(
+        source: &[T],
+        output: &mut bitvec::BitVec<bitvec::Msb0, u8>,
+        (size, ctx): (Size, Ctx),
+    ) -> Result<(), DekuError> {
+        dbg!(&output);
```
sharksforarms, line 102: nit: leftover `dbg!`

sharksforarms (commented):
This looks like a great first pass!

There's some clippy/build warnings mostly related to documentation. Docs will be important so users know how to use this awesome feature.

sharksforarms:
Hey @coolreader18 just a friendly ping! Are you still interested in working on this?

Issue #213, opened 2021-05-15
Request for Enhancement: Support for codepage conversion


v1gnesh:
Please consider adding support for custom encodings.
In https://github.com/sharksforarms/deku/issues/212, what I've described is the conversion of bytes from one encoding into String.
In my use case, there will be loads of `struct`s with loads of fields that I'll need to convert to String.
The `morph` referred there is essentially a codepage conversion function.

Like `#[deku(endian="big")]`, if codepage can be added as a top-level attribute, that'll save me loads & loads of LoC.
That is, if the codepage conversion can optionally be applied to all `struct` fields that are `String`.
Same as `endian`, settable at the top-level, field-level, etc.

Thanks again for this excellent library and all the work from the contributors!

sharksforarms:
Understood, this does seem a little tedious. Not sure I'm in favor of a default map function, it would only be able to accept a single type? What happens to the `a` field in this case?

Have you considered the alternative of splitting it out into a separate struct to hide some of these details?

```rust
use deku::prelude::*;
use std::convert::TryFrom;

#[derive(Debug, DekuRead)]
pub struct H {
  a: u16,
  #[deku(bytes = "4")]
  b: MorphedString,
}

#[derive(Debug, DekuRead)]
#[deku(ctx = "bytes: deku::ctx::Size")]
pub struct MorphedString(#[deku(map = "morph", bytes_read = "bytes.byte_size().unwrap()")] String);

pub fn morph(bytes: &[u8]) -> Result<String, DekuError> {
    let str = std::str::from_utf8(&bytes).unwrap().to_string();
    Ok(str)
}

fn main() {
    let input = b"\xAB\xCDDEKU";
    let the_h = H::try_from(input.as_ref()).unwrap();

    dbg!(&the_h);
    assert_eq!(the_h.a, 0xCDAB);
    assert_eq!(the_h.b.0, "DEKU");
}
```

You could implement Deref, PartialEq, etc. to make this type more transparent.

Possibly related: https://github.com/sharksforarms/deku/issues/174

v1gnesh:
> Not sure I'm in favor of a default map function, it would only be able to accept a single type?  
  
Yes, ideally configurable in a top-level setting such as `#[deku(cp_conv="morph", types="String")]`
  
> What happens to the a field in this case?
  
That'll be processed as normal; in this case, `#deku(endian="little")`

It will be really excellent to not have to code the `bytes_read` or any tagging for each of the `String` (in this case) fields.
[...]

PR #229, opened 2021-06-22
feat: Add `ctx_nested` attribute to top level

sharksforarms:
This allows clients to pass context arguments down to all fields.

Example
```rust
use deku::prelude::*;

struct MyCtx {}

#[deku(ctx = "ctx: MyCtx", ctx_nested = "ctx")]
pub struct Child {
    data: u32,
}

#[deku(ctx = "ctx: MyCtx", ctx_nested = "ctx")]
pub struct Parent {
    child: Child,
}

impl<'a> DekuRead<'a, MyCtx> for u32 {
    fn read(
        input: &'a deku::bitvec::BitSlice<deku::bitvec::Msb0, u8>,
        ctx: MyCtx,
    ) -> Result<(&'a deku::bitvec::BitSlice<deku::bitvec::Msb0, u8>, Self), DekuError>
    where
        Self: Sized,
    {
        // Custom u32 reader
        todo!()
    }
}

impl DekuWrite<MyCtx> for u32 {
    fn write(
        &self,
        output: &mut deku::bitvec::BitVec<deku::bitvec::Msb0, u8>,
        ctx: MyCtx,
    ) -> Result<(), DekuError> {
        // Custom u32 writer
        todo!()
    }
}
```

### DIFF
--- a/deku-derive/src/macros/mod.rs
+++ b/deku-derive/src/macros/mod.rs
@@ -235,18 +235,26 @@ fn gen_field_args(
     bits: Option<&Num>,
     bytes: Option<&Num>,
     ctx: Option<&Punctuated<syn::Expr, syn::token::Comma>>,
+    ctx_nested: Option<&Punctuated<syn::Expr, syn::token::Comma>>,
 ) -> syn::Result<TokenStream> {
     let crate_ = get_crate_name();
     let endian = endian.map(gen_endian_from_str).transpose()?;
     let bits = bits.map(|n| quote! {::#crate_::ctx::Size::Bits(#n)});
     let bytes = bytes.map(|n| quote! {::#crate_::ctx::Size::Bytes(#n)});
     let ctx = ctx.map(|c| quote! {#c});
+    let ctx_nested = ctx_nested.map(|c| quote! {#c});
 
     // FIXME: Should be `into_iter` here, see https://github.com/rust-lang/rust/issues/66145.
-    let field_args = [endian.as_ref(), bits.as_ref(), bytes.as_ref(), ctx.as_ref()]
-        .iter()
-        .filter_map(|i| *i)
-        .collect::<Vec<_>>();
+    let field_args = [
+        endian.as_ref(),
+        bits.as_ref(),
+        bytes.as_ref(),
+        ctx.as_ref(),
+        ctx_nested.as_ref(),
+    ]
+    .iter()
+    .filter_map(|i| *i)
+    .collect::<Vec<_>>();
 
     // Because `impl DekuRead<'_, (T1, T2)>` but `impl DekuRead<'_, T1>`(not tuple)
     match &field_args[..] {
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -100,7 +100,7 @@ fn cerror(span: proc_macro2::Span, msg: &str) -> TokenStream {
     syn::Error::new(span, msg).to_compile_error()
 }
 
-/// A post-processed version of `DekuReceiver`
+/// A post-processed version of [DekuReceiver](self::DekuReceiver)
 #[derive(Debug)]
 struct DekuData {
     vis: syn::Visibility,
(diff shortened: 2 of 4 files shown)


sharksforarms:
@RReverser @v1gnesh @constfold @wcampbell0x2a We can continue the discussion here

deku-derive/src/macros/mod.rs
```diff
@@ -235,18 +235,26 @@ fn gen_field_args(
...
-    let field_args = [endian.as_ref(), bits.as_ref(), bytes.as_ref(), ctx.as_ref()]
-        .iter()
-        .filter_map(|i| *i)
-        .collect::<Vec<_>>();
+    let field_args = [
+        endian.as_ref(),
+        bits.as_ref(),
+        bytes.as_ref(),
+        ctx.as_ref(),
+        ctx_nested.as_ref(),
```
sharksforarms, line 253: should be first?

wcampbell0x2a: I like `MyCtx` being first, since it's "non-default" deku usage and something you want to see first in impl error.

v1gnesh:
Apologies in advance if what I'm about to say makes little sense. I'm a pretty basic programmer...
If you have time, can you please help me understand how I can use this in the context of this comment - 
https://github.com/sharksforarms/deku/issues/213#issuecomment-841948929

So if I have a `struct` like this:
```Rust
#[deku(endian="little")]
pub struct H {
  a: u16,
  #[deku(map="morph", bytes_read="4")]
  b: String,
  #[deku(map="morph", bytes_read="4")]
  c: String,
  #[deku(map="morph", bytes_read="4")]
  d: String,
  #[deku(map="morph", bytes_read="64")]
  e: String,
}
```

I can now do this as:

```Rust
struct MyCtx {
  boits: [4, 4, 4, 64]
  funcs: ["morph", "morph", "morph", "morph"]
}

#[deku(ctx = "ctx: MyCtx", ctx_nested = "ctx")]
#[deku(endian="little")]
pub struct H {
  a: u16,
  b: String,
  c: String,
  d: String,
  e: String,
}
```

and 

```Rust
impl<'a> DekuRead<'a, MyCtx> for String {
    fn read(
        input: &'a deku::bitvec::BitSlice<deku::bitvec::Msb0, u8>,
        ctx: MyCtx,
    ) -> Result<(&'a deku::bitvec::BitSlice<deku::bitvec::Msb0, u8>, Self), DekuError>
    where
        Self: Sized,
    {
        // Custom String reader
        // interate/zip through ctx.boits and ctx.funcs and apply that to the `input`
        // ... what would the above impl look like?
    }
}
```

RReverser:
> @RReverser @v1gnesh @constfold @wcampbell0x2a We can continue the discussion here. Would a `ctx_nested` attribute be sufficient?

Yeah looks exactly like what I had in mind in https://github.com/sharksforarms/deku/issues/225#issuecomment-864578411:

> It's not very complicated to write one, but I wonder if it would make sense to add ctx_nested to Deku itself.

Seems sufficient for defining custom type structure. I wonder though, how does it interact with other ctx? Is it possible to still add extra `ctx` on fields and pass them down or would they conflict with `ctx_nested` on the struct itself?

constfold:
> Any thoughts on #225 (comment) ?

I don't like this idea at my first feeling. Changing the beheviour of build-in types is somewhat strange for me. But it is indeed a solution. If I must choose one between this and the `as`/`with`, I will vote for the latter one.
>  Does an attribute named ctx_nested make sense?

Maybe? Maybe `ctx_for_all`?

sharksforarms:
> https://github.com/sharksforarms/deku/issues/225#issuecomment-864477322 I think I'd like something even more general though - ideally I'd want to be able to define custom parsers for built-in types (e.g. LEB128 for integers, custom parser for floats, etc.) somewhere in a single place, and then somehow tell Deku to use those implementations for parsing, rather than putting `as`/`with` on each single field. I don't have a good idea in mind for how this would work though... Maybe some sort of context object to pass around?

@RReverser Is there a reason why you wouldn't want to abstract this into a separate type?

```rust
struct Leb128 {
    // ...
}

impl DekuRead for Leb128 { ... }
impl DekuWrite for Leb128 { ... }

struct ParseMe {
    field1: Leb128,
    // ...
}
```

RReverser:
> @RReverser Is there a reason why you wouldn't want to abstract this into a separate type?

Leb128 can be used for variety of types (u8, u16, u32, usize, i8, ...) and while they all share same encoding method, they still need to be reflected as distinct types in the final struct.

For me, the whole point of using a custom derive (whether the one I use in wasmbin or the one deku provides) is being able to use "natural" types in the final representation.

Thanks for the examples, I think this `ctx_nested` solution reduces the boilerplate significantly for those cases!

PR #235, opened 2021-09-25
Add read/write implems for generic arrays

xlambein:
Hey!  Thanks for the great project :-)

I needed to (de)serialize arrays of structs that implement `DekuRead`/`Write`, so I went ahead and added those trait implementations.  They also supersede the implementations for all the `uX`/`iX`/`fX` array/slices, so I removed those.

There's a bit of `unsafe` code to avoid doing unnecessary allocations.

### DIFF
--- a/src/impls/slice.rs
+++ b/src/impls/slice.rs
@@ -167,61 +167,57 @@ mod pre_const_generics_impl {
 mod const_generics_impl {
     use super::*;
 
-    macro_rules! ImplDekuSliceTraits {
-        ($($typ:ty),+ $(,)?) => {
-            $(
-                impl<Ctx: Copy> DekuWrite<Ctx> for &[$typ]
-                where
-                    $typ: DekuWrite<Ctx>,
-                {
-                    fn write(&self, output: &mut BitVec<Msb0, u8>, ctx: Ctx) -> Result<(), DekuError> {
-                        for v in *self {
-                            v.write(output, ctx)?;
-                        }
-                        Ok(())
-                    }
-                }
+    use std::mem::MaybeUninit;
 
+    impl<'a, Ctx: Copy, T, const N: usize> DekuRead<'a, Ctx> for [T; N]
+    where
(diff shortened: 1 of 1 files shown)


src/impls/slice.rs
```diff
@@ -167,61 +167,57 @@ mod pre_const_generics_impl {
...
+            input: &'a BitSlice<Msb0, u8>,
+            ctx: Ctx,
+        ) -> Result<(&'a BitSlice<Msb0, u8>, Self), DekuError>
+        where
+            Self: Sized,
+        {
+            #[allow(clippy::uninit_assumed_init)]
+            // This is safe because we initialize the array immediately after,
+            // and never return it in case of error
+            let mut slice: [T; N] = unsafe { MaybeUninit::uninit().assume_init() };
```
sharksforarms, line 186: The reason for this is because of potential performance impacts of default'ing here? `let mut slice: [$typ; N] = [<$typ>::default(); N];`

xlambein: Using `default()` requires that `T` implements `Default`, which imo is an unnecessary restriction.  To me the safe alternative would be to do:

```rust
let mut v = Vec::with_capacity(N);
let mut rest = input;
for i = 0..N {
    let (new_rest, value) = T::read(rest, ctx)?;
    v.push(value);
    rest = new_rest;
}
Ok((rest, v.try_into())
```

And later to convert it to an array.  That makes a few unnecessary allocations and checks, which are avoided with the unsafe section.  It doesn't seem very critical, so if you'd rather do that I think it's fine! :-)

sharksforarms:
Hey @xlambein sorry this took so long, this PR looks good to me.

xlambein:
Oh, I missed the previous comment.  Thanks a lot for merging!  And no worries about the delay, thanks for maintaining this project!

wcampbell0x2a:
Tracking issue: https://github.com/rust-lang/rust/issues/89379

PR #238, opened 2021-10-16
Sign extend signed integers

### LINKED ISSUE #236: Sign-extending signed integer types?
Currently, `deku` deserializes signed integers with limited bit width like unsigned integers, as evidenced by the following (passing) test:

```rust
use deku::{bitvec::BitSlice, ctx::Size, DekuRead as _};

#[test]
fn sign_extending() {
    let bits = &BitSlice::from_slice(&[0x1Fu8]).unwrap()[3..];
    let (_rest, num) = i16::read(bits, Size::Bits(5)).unwrap();

    assert_eq!(num, 31);
}
```

I would expect `num` to be `-1` instead, similar to how bit fields in C behave ([Compiler Explorer example](https://godbolt.org/z/c8Mjjf8c8)).

Is this something that could be supported? Perhaps using another context argument?

korrat:
This PR changes the implementation of `DekuRead` for signed integer primitive types, so that reads with a bit size less than the size of the underlying type are sign extended.

Fixes #236.

### DIFF
--- a/src/impls/primitive.rs
+++ b/src/impls/primitive.rs
@@ -97,6 +97,101 @@ macro_rules! ImplDekuRead {
     };
 }
 
+macro_rules! ImplDekuReadSignExtend {
+    ($typ:ty) => {
+        impl DekuRead<'_, (Endian, Size)> for $typ {
+            fn read(
+                input: &BitSlice<Msb0, u8>,
+                (endian, size): (Endian, Size),
+            ) -> Result<(&BitSlice<Msb0, u8>, Self), DekuError> {
+                let max_type_bits: usize = Size::of::<$typ>().bit_size();
+                let bit_size: usize = size.bit_size();
+
+                let input_is_le = endian.is_le();
+
+                if bit_size > max_type_bits {
+                    return Err(DekuError::Parse(format!(
+                        "too much data: container of {} bits cannot hold {} bits",
+                        max_type_bits, bit_size
+                    )));
+                }
+
+                if input.len() < bit_size {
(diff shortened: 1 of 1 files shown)


wcampbell0x2a:
Could you add tests to the `tests` directory? Just for proving these fields in structs and such.

sharksforarms:
> 
> 
> Could you add tests to the `tests` directory? Just for proving these fields in structs and such.

Since there's no change to attributes I think it's alright as is (current test cases were updated)

sharksforarms:
Friendly ping @korrat are you still interested in pursuing this PR?

korrat:
Sorry, @sharksforarms, totally forgot about this one. I'm still interested in pursuing this PR.

> Is there a way to reduce the duplication of code between signed/non-signed? I'm open to having this as a separate issue.

I've tried to merge the `ImplDekuRead` macros, but I'm having some trouble with the `f32`/`f64`, since they don't implement `Shl`/`Shr`. I can try again later in the week, but I think I would have to add a new macro for the floats. But perhaps that macro could be simpler.

korrat:
I can reproduce the compile error on master, so I think it is unrelated to the changes in this PR.

I think the error is related to rust-lang/rust#79609. Defining `_Unwind_Resume`, as suggested there, fixes the problem. Should I add a commit for the fix? Or would you prefer a different PR for that?

sharksforarms:
Hey @korrat sorry for my delay. The PR looks good at first glance, I'll take some time this week to look at it. A separate PR would be appreciated!

sharksforarms (approved):

sharksforarms:
Thanks! Released in 0.13.0

PR #239, opened 2021-11-05
Show struct ident in Assertion error message

wcampbell0x2a:
This allows quicker debugging and more information to user of library when an assertion fails.


sharksforarms:
Thanks!

PR #247, opened 2022-02-23
Fix link error in ensure_no_std

korrat:
On some targets, `alloc` is built with unwinding. Therefore, it requires `libunwind` even when compiling with `panic=abort`. Defining the `_Unwind_Resume` symbol fixes this. The actual implementation does not matter, since it will never be called.

See rust-lang/rust#79609 for more details.


sharksforarms:
Thanks! Released in 0.13.0

PR #255, opened 2022-04-06
DekuError: Derive Clone

interruptinuse:
Would be nice to have, but not pushing for it (in case there are plans to propagate a `Box<dyn Error>` or something like that).


sharksforarms (approved):
Thanks!

Issue #168, opened 2021-01-13
Add debug attribute


wcampbell0x2a:
It would be helpful to have a `debug` attribute for easier debugging the decode/encode process. It would do the following:
- Display the bits before reading in T (`.view_bits()`)
- Display result of T
- Display the bits after reading in T (`.view_bits()`)

I *think* this could be implemented for both `field` and `top-level` attributes.

sharksforarms:
Yes, there needs to be a better way to debug parsers. I don't mind the idea, I think a mix of logging throughout the crate and a debug attribute like you mentioned could be used. This could be feature gated, to avoid unnecessary checks.

wcampbell0x2a:
I've never used any logging crates,  do you have one in mind?

sharksforarms:
https://github.com/rust-lang/log#in-libraries

wcampbell0x2a:
Our usage wouldn't technically be `in-libraries`, since it's mostly emitted code.

wcampbell0x2a:
Closing, the feature `logging` works for this

PR #271, opened 2022-07-04
Add logging feature

wcampbell0x2a:
* Improve ease of debugging an incorrect parser implementation by adding
  context to the Err values from deku parsers by adding logging from the
  log crate.
* This is feature-gated under the new "logging" feature.

From the following sample program, you can easily see the improved error
message and debugging since we known the field that was incorrect
parsed. Much easier than gdb.
```
[dependencies]
deku = { path = "../deku", features = ["logging"]}
log = "0.4.17"
env_logger = "0.9.0"
```
```
use deku::prelude::*;

pub struct TestStruct {
    pub a: u16,
    pub b: u16,
}

fn main() {
    env_logger::init();
    TestStruct::from_bytes((&[0x01, 0x02, 0x03], 0)).unwrap();
    println!("Hello, world!");
}
```
```
> RUST_LOG=trace cargo r
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/deku-testing`
[2022-07-04T16:11:54Z TRACE deku_testing] Reading: TestStruct::a from [00000001, 00000010, 00000011]
[2022-07-04T16:11:54Z TRACE deku_testing] Reading: TestStruct::b from [00000011]
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Incomplete(NeedSize { bits: 16 })', src/main.rs:11:54
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This could be expanded in the future to enums and/or other internal deku
structs.

Even cooler would be something like this:
```text
                                                                      |-----------------|
[2022-07-04T16:08:16Z TRACE deku_testing] Reading: TestStruct::a from [00000001, 00000010, 00000011]
```

See #168

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ members = [
 default = ["std", "const_generics"]
 std = ["deku_derive/std", "bitvec/std", "alloc"]
 alloc = ["bitvec/alloc"]
+logging = ["deku_derive/log"]
 const_generics = []
 
 [dependencies]
--- a/deku-derive/Cargo.toml
+++ b/deku-derive/Cargo.toml
@@ -13,6 +13,7 @@ proc-macro = true
 
 [features]
 std = ["proc-macro-crate"]
+logging = ["log"]
 
 [dependencies]
 quote = "1.0"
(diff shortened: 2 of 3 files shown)


sharksforarms:
LGTM thanks!

I think this format from @mycoliza could be really cool for debugging: https://twitter.com/mycoliza/status/1529987746269233152/photo/2

sharksforarms:
@wcampbell0x2a https://crates.io/crates/mycelium-bitfield just dropped 👀

Issue #44, opened 2020-06-04
Differing impls for bits and bytes


sharksforarms:
If `bytes` attribute is used and the index is on a byte boundary, it may be quicker to read from a slice of &[u8] instead of reading 8*n bits.

I'd like for more benchmarks to be written before so this optimization can be measured

One option could be to feature flag the bits/bytes attributes

constfold:
Since #61 has been merged, this could be done by:
1. Add a new type `ByteSize`
2. Add specialization like `impl DekuRead<ByteSize> for xxx`
3. bytes uses type `ByteSize`, bits uses type `BitSize`, then compiler will decide which impl should be called.

sharksforarms:
Great idea, I'd like to see a better suite of benchmarks before this is tackled so we can measure the difference

https://github.com/sharksforarms/deku/issues/25

wcampbell0x2a:
Closing, the reader does this

Issue #25, opened 2020-05-28
Add benchmarks


sharksforarms:
Using criterion

+bonus if compared to other crates like `nom` for reading

sharksforarms:
Added some simple benchmarks to catch regressions and measure optimizaitons

OliverGavin:
I was playing with the magic attribute and benchmarks and I realized it is very slow so I thought I would share:

```RUST
#![feature(test)]

use deku::prelude::*;

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
#[deku(endian = "big", magic = b"\x00\x00\x00\x05")]
struct WithMagicV5 {
    value: u32
}

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
#[deku(endian = "big")]
struct WithoutMagicV5 {
    #[deku(assert = "*version == 5")]
    version: u32,  // users ideally should not be able to set this as it is always 5
    value: u32
}

#[cfg_attr(test, macro_use)]
extern crate hex_literal;

#[cfg(test)]
mod magic_benches {
    use super::*;

    extern crate test;
    use test::Bencher;
    
    #[bench]
    fn bench_with_magic(b: &mut Bencher) {
        let data = hex!("00 00 00 05 00 00 00 01");
        b.iter(|| {
            let (_, _) = WithMagicV5::from_bytes((&data, 0)).unwrap();
        });
    }
    
    #[bench]
    fn bench_without_magic(b: &mut Bencher) {
        let data = hex!("00 00 00 05 00 00 00 01");
        b.iter(|| {
            let (_, _) = WithoutMagicV5::from_bytes((&data, 0)).unwrap();
        });
    }
}
```

And the results:
```
running 2 tests
test magic_benches::bench_with_magic        ... bench:          48 ns/iter (+/- 4)
test magic_benches::bench_without_magic     ... bench:          21 ns/iter (+/- 6)
```

sharksforarms:
> 
> I was playing with the magic attribute and benchmarks and I realized it is very slow so I thought I would share:
> 

Nice one!

There are some optimizations which could be done. This is a great example.

For example, `magic` attribute handling could convert the byte string to `0x00000005u32` (at compile time?) and do a single `u32` read and equality check. Currently, it would do 4 `u8` reads and 4 equality checks (iterating the byte string).

OliverGavin:
[...]

Issue #224, opened 2021-06-10
Panic: error parsing from slice: could not convert slice to array


wcampbell0x2a:
Adding the following example in deku causes this error:
```
---- it_works stdout ----
thread 'it_works' panicked at 'called `Result::unwrap()` on an `Err` value: Parse("error parsing from slice: could not convert slice to array")', tests/test_struct.rs:228:50
```

```diff
diff --git a/tests/test_struct.rs b/tests/test_struct.rs
index 062d006..66c015a 100644
--- a/tests/test_struct.rs
+++ b/tests/test_struct.rs
@@ -167,3 +167,64 @@ fn test_raw_identifiers_struct() {
     let ret_write: Vec<u8> = ret_read.try_into().unwrap();
     assert_eq!(test_data, ret_write);
 }
+
+#[test]
+fn it_works() {
+    #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+    pub struct Packet {
+        pub data: Data,
+    }
+
+    #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+    pub struct Data {
+        pub one: One,
+        pub two: Two,
+    }
+
+    #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+    #[deku(type = "u8", bits = "2")]
+    pub enum One {
+        Start = 0,
+        Go = 1,
+        Stop = 2,
+    }
+
+    #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+    #[deku(type = "u8", bits = "4")]
+    pub enum Two {
+        #[deku(id = "0b0000")]
+        Put(Op1),
+        #[deku(id = "0b0001")]
+        Store(Op1),
+        #[deku(id = "0b0010")]
+        Load(Op2),
+        #[deku(id = "0b0011")]
+        Allocate(Op2),
+    }
+
+    #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+    pub struct Op1 {
+        #[deku(bits = "2")]
+        pub i: u8,
+        #[deku(bits = "4")]
+        pub o: u8,
+    }
+
+    #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+    pub struct Op2 {
+        #[deku(bits = "8")]
+        pub w: u8,
+        #[deku(bits = "6")]
+        pub j: u8,
+    }
+
+    let mut packet = Packet {
+        data: Data {
+            one: One::Start,
[...]

PR #278, opened 2022-09-28
Specialize bytes

wcampbell0x2a:
Remove Size in favor of a BitSize and ByteSize. This allows the
Deku{Read/Write}(Endian, BitSize) and Deku{Read/Write}(Endian, ByteSize)
impls to be created and allow the ByteSize impls to be faster. Most of
the assumption I made (and the perf that is gained) is from the removal
of this branch for bytes:

```rust
let value = if pad == 0
                      && bit_slice.len() == max_type_bits
                      && bit_slice.as_raw_slice().len() * 8 == max_type_bits
                  {
```

See https://github.com/sharksforarms/deku/issues/44

I Added some benchmarks, in order to get a good idea of what perf this
allows. The benchmarks look pretty good, but I'm on my old thinkpad.
These results are vs the benchmarks running from master.

```
deku_read_vec           time:   [744.39 ns 751.13 ns 759.58 ns]
                        change: [-23.681% -21.358% -19.142%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe
```

 See https://github.com/sharksforarms/deku/issues/25

### DIFF
--- a/src/impls/primitive.rs
+++ b/src/impls/primitive.rs
@@ -318,9 +318,9 @@ macro_rules! ForwardDekuRead {
 
                 // Since we don't have a #[bits] or [bytes], check if we can use bytes for perf
                 if (bit_size.0 % 8) == 0 {
-                    <$typ>::read(input, (endian, bit_size))
-                } else {
                     <$typ>::read(input, (endian, ByteSize(bit_size.0 / 8)))
+                } else {
+                    <$typ>::read(input, (endian, bit_size))
                 }
             }
         }
--- a/src/impls/primitive.rs
+++ b/src/impls/primitive.rs
@@ -5,8 +5,122 @@ use core::convert::TryInto;
 #[cfg(feature = "alloc")]
 use alloc::format;
 
+// TODO: duplicate code, separate macros for bits vs bytes? we only want to specialize reading a byte u8...
+impl DekuRead<'_, (Endian, BitSize)> for u8 {
+    fn read(
+        input: &BitSlice<Msb0, u8>,
+        (endian, size): (Endian, BitSize),
+        ) -> Result<(&BitSlice<Msb0, u8>, Self), DekuError> {
+        let max_type_bits: usize = BitSize::of::<u8>().0;
+        let bit_size: usize = size.0;
+
+        let input_is_le = endian.is_le();
+
+        if bit_size > max_type_bits {
+            return Err(DekuError::Parse(format!(
+                        "too much data: container of {} bits cannot hold {} bits",
+                        max_type_bits, bit_size
+                        )));
+        }
+
+        if input.len() < bit_size {
+            return Err(DekuError::Incomplete(crate::error::NeedSize::new(bit_size)));
+        }
+
+        let (bit_slice, rest) = input.split_at(bit_size);
+
+        let pad = 8 * ((bit_slice.len() + 7) / 8) - bit_slice.len();
+
+        let value = if pad == 0
+            && bit_slice.len() == max_type_bits
+            && bit_slice.as_raw_slice().len() * 8 == max_type_bits
+            {
+                // if everything is aligned, just read the value
+                let bytes: &[u8] = bit_slice.as_raw_slice();
+
+                // Read value
+                if input_is_le {
+                    <u8>::from_le_bytes(bytes.try_into()?)
+                } else {
+                    <u8>::from_be_bytes(bytes.try_into()?)
+                }
+            } else {
+                // Create a new BitVec from the slice and pad un-aligned chunks
+                // i.e. [10010110, 1110] -> [10010110, 00001110]
+                let bits: BitVec<Msb0, u8> = {
+                    let mut bits = BitVec::with_capacity(bit_slice.len() + pad);
+
+                    // Copy bits to new BitVec
+                    bits.extend_from_bitslice(bit_slice);
+
+                    // Force align
+                    //i.e. [1110, 10010110] -> [11101001, 0110]
+                    bits.force_align();
+
+                    // Some padding to next byte
+                    let index = if input_is_le {
+                        bits.len() - (8 - pad)
+                    } else {
+                        0
+                    };
+                    for _ in 0..pad {
+                        bits.insert(index, false);
+                    }
+
+                    // Pad up-to size of type
+                    for _ in 0..(max_type_bits - bits.len()) {
+                        if input_is_le {
+                            bits.push(false);
+                        } else {
+                            bits.insert(0, false);
+                        }
+                    }
+
+                    bits
+                };
+
+                let bytes: &[u8] = bits.as_raw_slice();
+
+                // Read value
+                if input_is_le {
+                    <u8>::from_le_bytes(bytes.try_into()?)
+                } else {
+                    <u8>::from_be_bytes(bytes.try_into()?)
+                }
+            };
+        Ok((rest, value))
+    }
+}
+
+// specialize u8
+impl DekuRead<'_, (Endian, ByteSize)> for u8 {
+    fn read(
+        input: &BitSlice<Msb0, u8>,
+        (_, size): (Endian, ByteSize),
+        ) -> Result<(&BitSlice<Msb0, u8>, Self), DekuError> {
+        let max_type_bits: usize = BitSize::of::<u8>().0;
+        let bit_size: usize = size.0 * 8;
+
+        if bit_size > max_type_bits {
+            return Err(DekuError::Parse(format!(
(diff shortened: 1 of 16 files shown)


wcampbell0x2a:
I'm getting errors on the following lines:
```
failures:
    src/attributes.rs - attributes (line 1130)
    src/attributes.rs - attributes (line 270)
```
For these tests, not sure why:
 Example:
  ```rust
  # use deku::prelude::*;
  # use std::convert::{TryInto, TryFrom};
  # #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
  #[deku(type = "u32", bytes = "2")]
  enum DekuTest {
      #[deku(id = "0xBEEF")]
      VariantA(u8),
  }

  let data: Vec<u8> = vec![0xEF, 0xBE, 0xFF];

  let value = DekuTest::try_from(data.as_ref()).unwrap();

  assert_eq!(
      DekuTest::VariantA(0xFF),
      value
  );

  let value: Vec<u8> = value.try_into().unwrap();
  assert_eq!(data, value);
  ```

Example:
  ```rust
  # use deku::prelude::*;
  # use std::convert::{TryInto, TryFrom};
  # #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
  struct DekuTest {
      #[deku(bytes = 2)]
      field_a: u32,
      field_b: u8, // defaults to size_of<u8>
  }

  let data: Vec<u8> = vec![0xAB, 0xCD, 0xFF];

  let value = DekuTest::try_from(data.as_ref()).unwrap();

  assert_eq!(
      DekuTest {
         field_a: 0xCDAB,
         field_b: 0xFF,
      },
      value
  );

  let value: Vec<u8> = value.try_into().unwrap();
  assert_eq!(data, value);
```

**Comment by wcampbell0x2a** [comment]
@sharksforarms @constfold lmk what you think

File: src/impls/primitive.rs [hunk 1]
```diff
@@ -90,7 +89,41 @@ macro_rules! ImplDekuRead {
...
+                    )));
+                }
+
+                if input.len() < bit_size {
+                    return Err(DekuError::Incomplete(crate::error::NeedSize::new(bit_size)));
+                }
+
+                let (bit_slice, rest) = input.split_at(bit_size);
+
+                let bytes: &[u8] = bit_slice.as_raw_slice();
```
constfold, line 119: This is exactly what I want in that comment. But... 
> the removal of this branch for bytes
> ```rust
> if pad == 0
> && bit_slice.len() == max_type_bits
> && bit_slice.as_raw_slice().len() * 8 == max_type_bits
> ```
Can we really do this? What if `bit_slice` is not byte aligned?

constfold:
> ```
> failures:
>    src/attributes.rs - attributes (line 1130)
>    src/attributes.rs - attributes (line 270)
> ```
two error messages both said:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Parse("error parsing from slice: could not convert slice to array")', src/attributes.rs:15:47
```
It looks like #224, caused by alignment issue(might be the one I mentioned above).

src/impls/primitive.rs
sharksforarms: We should be good given:
- `let bit_size: usize = size.0 * 8;` -- will always be byte aligned
- we check `bit_size > max_type_bits` & `input.len() < bit_size` before calling `split_at`

sharksforarms:
Hey @wcampbell0x2a, nice work! This is looking great (thanks for the benchmarks!)

I don't have any comments apart from the test cases, I can offer my understanding on why they're failing.

> ```rust
> #[deku(type = "u32", bytes = "2")]
> enum DekuTest {
>     #[deku(id = "0xBEEF")]
>     VariantA(u8),
> }
>
> let data: Vec<u8> = vec![0xEF, 0xBE, 0xFF];
> let value = DekuTest::try_from(data.as_ref()).unwrap();
> ```


So, the issue with these tests is that the changes no longer consider padding when reading:

Original code:
```rust
                let value = if pad == 0
                    && bit_slice.len() == max_type_bits
                    && bit_slice.as_raw_slice().len() * 8 == max_type_bits
                {
                    // if everything is aligned, just read the value
```

This section of code is an optimisation for when we have *exactly what we want*.

The above test using this PR, will read 2 bytes and try to convert that slice to a u32 and fail (u32 needs 4 bytes).

To fix this, we'd need to bring padding back some way or another. I'm thinking it may be possible to pad using arithmetic instead of allocation too. Maybe a later optimisation.

wcampbell0x2a:
> So, the issue with these tests is that the changes no longer consider padding when reading:

Could we add a `Padding` struct for this? When `bytes` is more than `type`? Are there other cases in which this would happen for a field in a struct as well as this enum usage? This also wouldn't happen  for `type = u8` as I see it.

Just trying to reduce as many branches as I can for performance.

wcampbell0x2a:
Current benchmarking :chart_with_upwards_trend: 
```
deku_read_byte          time:   [8.3902 ns 8.4387 ns 8.4959 ns]
                        change: [-17.300% -16.110% -14.983%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 18 outliers among 100 measurements (18.00%)
  1 (1.00%) low mild
  8 (8.00%) high mild
  9 (9.00%) high severe

deku_write_byte         time:   [63.329 ns 63.648 ns 64.012 ns]
                        change: [-7.6643% -5.4937% -3.9645%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 23 outliers among 100 measurements (23.00%)
  1 (1.00%) low severe
  8 (8.00%) low mild
  6 (6.00%) high mild
  8 (8.00%) high severe

deku_read_bits          time:   [500.82 ns 504.55 ns 509.19 ns]
                        change: [-0.9629% +1.7699% +4.4820%] (p = 0.21 > 0.05)
                        No change in performance detected.
Found 18 outliers among 100 measurements (18.00%)
  4 (4.00%) high mild
  14 (14.00%) high severe

deku_write_bits         time:   [95.921 ns 96.374 ns 96.879 ns]
                        change: [-11.428% -9.0292% -7.2123%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
  7 (7.00%) high severe

deku_read_enum          time:   [14.375 ns 14.418 ns 14.466 ns]
                        change: [-21.776% -19.593% -17.363%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  1 (1.00%) low mild
  7 (7.00%) high mild
  7 (7.00%) high severe

deku_write_enum         time:   [101.08 ns 101.55 ns 102.10 ns]
                        change: [-10.996% -8.9539% -7.4107%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 18 outliers among 100 measurements (18.00%)
  7 (7.00%) high mild
  11 (11.00%) high severe

deku_read_vec           time:   [777.98 ns 779.94 ns 782.09 ns]
                        change: [-13.747% -12.440% -10.565%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  1 (1.00%) low mild
  7 (7.00%) high mild
  2 (2.00%) high severe

deku_write_vec          time:   [4.2898 µs 4.3064 µs 4.3268 µs]
                        change: [-14.469% -12.650% -10.762%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 18 outliers among 100 measurements (18.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  12 (12.00%) high severe
  ```

**Comment by wcampbell0x2a** [comment]
One of the CI failures is from
```
warning: you are deriving `PartialEq` and can implement `Eq`
  --> src/ctx.rs:76:30
   |
76 | #[derive(Debug, Copy, Clone, PartialEq, Eq, Ord, PartialOrd)]
   |                              ^^^^^^^^^ help: consider deriving `Eq` as well: `PartialEq, Eq`
```
```
77: pub enum Limit<T, Predicate: FnMut(&T) -> bool> {
```
which looks to be a duplicate of: https://github.com/rust-lang/rust-clippy/issues/9413

sharksforarms:
Thanks! There was a couple more TODO i added in regards to code duplication that should be addressed before merging

wcampbell0x2a:
> Thanks! There was a couple more TODO i added in regards to code duplication that should be addressed before merging

I _think_ I fixed those now.

sharksforarms:
Thanks! Release pending :)

PR #283, opened 2022-10-08
Fix issue where endianness trickery wouldn't swap bytes correctly when reading less than the full amount into an integer

### LINKED ISSUE #282: `#[deku(bits = 24)]` and `#[deku(bytes = 3)]` have different behaviors reading big-endian `u32`
Here's a simplified example of my code:

```rust
use std::convert::TryFrom;
use deku::prelude::*;

#[deku(endian = "big")]
struct Foo {
    #[deku(bits = 24)]
    bits: u32,

    #[deku(bytes = 3)]
    bytes: u32,
}

#[test]
fn this_fails() {
    let data: &[u8] = &[0xAC, 0x1F, 0xBD, 0xAC, 0x1F, 0xBD];
    let foo: Foo = Foo::try_from(data).unwrap();
    assert_eq!(foo.bits, 0xAC1FBD);
    assert_eq!(foo.bytes, 0xAC1FBD); // Fails with a value of 0xAC1FBD00
}
```

Instead of placing two zeroes at the front of the integer, they get placed at the back.

caass:
- Add regression test
- add little-endian test as well
- Fix issue where endianness trickery wouldn't swap bytes correctly when reading less than the full amount

Closes #282

### DIFF
--- a/src/impls/primitive.rs
+++ b/src/impls/primitive.rs
@@ -168,17 +168,19 @@ macro_rules! ImplDekuReadBytes {
                     // cannot use from_X_bytes as we don't have enough bytes for $typ
                     // read manually
                     let mut res: $inner = 0;
-                    for b in bytes.iter().rev() {
-                        res <<= 8 as $inner;
-                        res |= *b as $inner;
-                    }
-
                     if input_is_le {
-                        res as $typ
+                        for b in bytes.iter().rev() {
+                            res <<= 8 as $inner;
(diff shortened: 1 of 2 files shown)


caass:
I see CI is failing, but it looks unrelated to my PR? Let me know if there's something I need to do to fix this :)

sharksforarms:
Looks good to me. Good catch! Thanks for the tests.

Issue #292, opened 2022-11-14
Inconsistent behaviour for u8 unaligned read when not explicitly setting bit size and/or endianness


calebfletcher:
When a u8 is read with a non-bit-aligned reader, the value is different depending on whether an explicit `#[deku(bits = 8)]` attribute is used and whether an explicit endianness is set.

When either the endianness is inherited from the machine or the `#[deku(bits = 8)]` is not used, the following code reads `field2` as 15. When both attributes are set, the value is 255. Note that the endianness can be set to either big or little, both produce the result 255, which I believe is the correct result in this case.

For a minimal reproducible example, run the following code with different combinations of the `#[deku(endian = "little")]` and `#[deku(bits = 8)]` attributes commented out.
```rust
use deku::prelude::*;

#[derive(Debug, PartialEq, DekuRead)]
#[deku(endian = "little")]
struct Container {
    #[deku(bits = 4)]
    field1: u8,
    #[deku(bits = 8)]
    field2: u8,
    #[deku(bits = 4)]
    field3: u8,
}

fn main() {
    let test_data: &[u8] = [0x0F, 0xF0].as_ref();
    dbg!(Container::from_bytes((test_data, 0)).unwrap());
}
```

Additionally, this behavior seems to have changed in commit 0d03d70. Prior to this commit (the previous commit was aeb9741), no matter whether endianness is explicit or implicit, the expected value of 255 was returned. Following this commit, however, letting the endianness be implicitly determined while specifying the bit size now returns 15 as well.

| commit | no endian, no bit size | little endian, no bit size | big endian, no bit size | no endian, bit size 8 | little endian, bit size 8 | big endian, bit size 8 |
| --- | --- | --- | --- | --- | --- | --- |
| aeb9741 | 15 | 15 | 15 | 255 | 255 | 255 |
| 0d03d70 |  15 | 15 | 15 |  15 | 255 | 255 |

In the cases when 15 is returned, the remaining fields are all zeros, which makes it seem as though the extra bits are getting dropped entirely. It still advances the reader the expected amount.

## Expected Behavior
The expected behavior, as I understand it, should be that specifying little endian should produce identical results to letting Deku infer the endianness from the machine's architecture on a little endian machine. Additionally, reading an unaligned u8 should produce identical results with and without a bit size of 8.

sharksforarms:
This is a bug, might have time to look at this today I have a feeling it's due to recent changes 

cc @wcampbell0x2a

[...]

PR #294, opened 2022-11-15
Fix regression with unaligned u8

wcampbell0x2a:
* Add logic to fix alignment when needed
* Add tests for regression testing

See #292


wcampbell0x2a:
Looks like the CI errors for miri occur on master also

calebfletcher:
Thanks for fixing the case for an offset of 4, but it seems like using an offset of 2 still causes it to fail:

```rust
let test_data: &[u8] = [0b11000000, 0b00111111].as_ref();

#[derive(Debug, PartialEq, DekuRead)]
#[deku(endian = "little")]
struct Container {
    #[deku(bits = 2)]
    field1: u8,
    field2: u8,
    #[deku(bits = 6)]
    field3: u8,
}

assert_eq!(
    Container::from_bytes((test_data, 0)).unwrap().1,
    Container {
        field1: 0b11,
        field2: 0,
        field3: 0b111111,
    }
);
```

Adding in an explicit `#[deku(bits = 8)]` to `field2` causes this case to work as well, so I believe it is likely related.

sharksforarms (approved):
Thanks @wcampbell0x2a, changes and tests look good to me.

PR #295, opened 2022-11-16
DRAFT: Add aligned attribute

wcampbell0x2a:
This is a draft for now, but I tested out adding an `aligned` attribute to test the performance increase.

```
deku_read_vec           time:   [765.59 ns 769.06 ns 772.51 ns]
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
deku_read_vec_perf      time:   [686.51 ns 691.37 ns 696.95 ns]
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe
```

not bad

@sharksforarms I'm curious of what you think of the approach of adding an attribute and ctx for this? I acknowledge it's a bit of a foot-gun.

### DIFF
--- a/benches/deku.rs
+++ b/benches/deku.rs
@@ -21,13 +21,13 @@ enum DekuEnum {
     VariantA(u8),
 }
 
-/// This is faster, because we go right to (endian, bytes)
+/// This is faster, because we go right to (endian, bytes, Aligned)
 #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
 struct DekuVecPerf {
-    #[deku(bytes = "1")]
+    #[deku(aligned)]
     count: u8,
     #[deku(count = "count")]
-    #[deku(bytes = "1")]
+    #[deku(aligned)]
     data: Vec<u8>,
 }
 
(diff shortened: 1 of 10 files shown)


sharksforarms:
Thanks for working on this.

I like the idea. I wonder if it we could assume types aligned by default, then certain attributes may invalidate this assumption such as `bits`, `reader`, ..., WDYT?

If I understand correctly, this is to avoid the un-aligned allocation path right? (We can confirm via a test in `test_alloc.rs`)

wcampbell0x2a:
> I like the idea. I wonder if it we could assume types aligned by default, then certain attributes may invalidate this assumption such as `bits`, `reader`, ..., WDYT?

I think this could be hard to track, as alignment in the parent struct could effect the alignment of the child struct. I added a test case in which case this causes a panic b/c of lack of alignment: https://github.com/sharksforarms/deku/pull/295/files#diff-d0c3d1ab0ac5b9134d620e8e6f9574d3f0545b36474f310912b2a85ad6ded372R37


I actually think in most cases, you would set the alignment maybe in the top-level(something I haven't implemented).
```rust
#[deku(endian = "big", aligned)]
struct DekuTest {
    field_a: u8,
    field_b: u8,
    field_c: u16,
}
```

PR #309, opened 2023-01-12
fix clippy warnings

caass:
I'm not sure how these got in here since there's CI to deny them, but it was blowing up my editor.


wcampbell0x2a (approved):

sharksforarms:
The CI only runs `clippy`, if we wanted to catch everything(?) we could do something like `cargo clippy --all-targets --tests --benches --workspace`

PR #315, opened 2023-01-19
Support multiple fields to `id`

wcampbell0x2a:
- Add use of multiple id's from ctx in enums by expanding the match
  statement to tuples

### DIFF
--- a/tests/test_attributes/test_ctx.rs
+++ b/tests/test_attributes/test_ctx.rs
@@ -116,16 +116,27 @@ fn test_struct_enum_ctx_id() {
     }
 
     #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
+    #[deku(ctx = "my_id: u8", id = "my_id")]
+    enum EnumJustId {
+        #[deku(id = "1")]
+        VarA(u8),
+        #[deku(id = "2")]
+        VarB,
+    }
+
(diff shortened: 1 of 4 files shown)


wcampbell0x2a:
`deku::rest` is currently not expanded for ctx, not sure how to fix this.
```
error[E0425]: cannot find value `__deku_rest` in this scope
  --> examples/enums.rs:20:22
   |
20 |         #[deku(ctx = "*my_id, deku::rest.len()")]
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
```

tests/test_attributes/test_ctx.rs
```diff
@@ -105,36 +105,72 @@ fn test_top_level_ctx_enum_default() {
 #[test]
 fn test_struct_enum_ctx_id() {
     #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
-    #[deku(ctx = "my_id: u8", id = "my_id")]
+    #[deku(ctx = "my_id: u8, data: usize", id = "my_id, data")]
```
sharksforarms, line 108: We might want to duplicate the test? (one tuple, one non-tuple)

sharksforarms (approved):
Thanks! 🙏

PR #322, opened 2023-02-28
Temp value writer

SneakyBerry:
Provide value for temp field on write.
IMO it is suitable for a lot of cases, e.g write vector with predicate of it's length, compute mask of option values and so on.


sharksforarms (approved):
Looks great. Thank you!

PR #331, opened 2023-04-07
Fix std::result::Result path in deku-derive

### LINKED ISSUE #310: Type mismatch when using anyhow
When I import `anyhow::{Result, Ok}`, a type mismatch with DekuError and anyhow::Error occurs in the derive macros for deku.

wcampbell0x2a:
Use std::result::Result in all derived quotes in deku-derive{read/write}.

Closes #310


sharksforarms (approved):
Thanks!

PR #334, opened 2023-05-04
Reduce calls to bivec.domain().region()

wcampbell0x2a:
This isn't as much as a performance gain as I thought it would be (around %2), but an improvement neverless.


sharksforarms (approved):
Thanks!

Issue #328, opened 2023-03-29
Enum discriminant re/mapping


IniterWorker:
Let's say you want to map an enum that's not the perfect mirror of the data.

Is there any way to do it with the current proc-macros?

## Example

```rust
pub enum TestDesMapping {
	/// A
	A = 0,
	/// B
	B = 1,
	/// C everything in range (C_desc + 1)..
	C = 2,
```

## Current behaviour

If we use `id_pat` or `id` like below then we will raise an error on `DekuRead` method  `try_from`  :  `Too much data`.

```rust
#[derive(DekuWrite, DekuRead)]
pub enum TestDesMapping {
	/// A
	A = 0,
	/// B
	B = 1,
	/// C everything after
	#[deku(id_pat = "_")]
	C = 2,
```

## Expected Behaviour

In this scenario, I should expect to have `C` for the range `(C_desc + 1)..`.

sharksforarms:
Can you use `id_pat` like so? https://github.com/sharksforarms/deku/blob/13fc4e9b084042ac77f9a3f6aa366f791ddac646/examples/enums.rs#L20-L21

Are you able to provide a minimal example to showcase what you're trying to do?

IniterWorker:
I would like to use a pattern match such as `id_pat`. But currently it doesn't fit my use case.

## Failures

```
failures:
    dao::types::probe_model::test2::test_basic_c
    dao::types::probe_model::test2::test_basic_pattern
```

## Example

```rust
#[cfg(test)]
mod test {
    use deku::prelude::*;

    #[derive(Clone, Copy, PartialEq, Eq, Debug, DekuWrite, DekuRead)]
    #[deku(type = "u8")]
    #[non_exhaustive]
    #[repr(u8)]
    pub enum TestDesMapping {
        /// A
        A = 0,
        /// B
        B = 1,
        /// C everything after
        #[deku(id_pat = "_")]
        C = 2,
    }

    #[test]
    fn test_basic_a() {
        let input = [0u8];
        let ret_read = TestDesMapping::try_from(input.as_slice()).unwrap();
        assert_eq!(TestDesMapping::A, ret_read);
        let ret_write: Vec<u8> = ret_read.try_into().unwrap();
        assert_eq!(input.to_vec(), ret_write);
    }

    #[test]
    fn test_basic_c() {
        let input = [2u8];
        let ret_read = TestDesMapping::try_from(input.as_slice()).unwrap();
[...]

PR #335, opened 2023-05-09
Add DekuRead's catch_all feature

IniterWorker:
Introducing the new `catch_all` feature in the derive crate for `DekuRead`!

New feature that allows developers to catch every value from a variant without reporting any `DekuError`. This is particularly useful in cases where unpredictable data may exist in an enum/protocol.

With the catch_all feature, you no longer need to worry about handling each possible value explicitly. Instead, you can catch all values with a single line of code. This simplifies the development process and saves valuable time.

I've taken care to thoroughly test this feature, with tests covering compilation and end-to-end integration. 

@sharksforarms, I value your input and would love to hear your thoughts on this new feature. Please don't hesitate to reach out with any feedback or suggestions for future improvements.

Will close https://github.com/sharksforarms/deku/issues/328


IniterWorker:
Hi @sharksforarms,

Do you wait for any improvement on my side ?

wcampbell0x2a:
> Hi sorry, life has been busy.
> 
> The change looks good to me, thanks for adding tests. I'm not sure of the name `catch_all`, I would prefer something like `default`, wdyt? cc @wcampbell0x2a

Sure! I think it more closely follows https://doc.rust-lang.org/std/default/trait.Default.html#enums (we could parse that in the future, or not idk).

sharksforarms (approved):
Looks good to me, thanks for taking the time to create this PR and adding tests!

PR #341, opened 2023-06-04
Add bound attribute to restrict derived impls

MegaBluejay:
Analogous to [serde's](https://serde.rs/attr-bound.html)

This allows using generics without restricting them in the struct definition

The use-case for me is parsing into a raw form, then post-processing some fields, while retaining most of the structure. With this both can use the same generic type, and only the raw form needs to implement DekuRead/DekuWrite.

Inferring the bounds automatically would be better, but I'm not up to writing that right now.


wcampbell0x2a (changes requested):
Could you add documentation to `src/attributes.rs`? Thanks

wcampbell0x2a:
Closed by #685

Issue #344, opened 2023-06-13
RFC: API breaking performance improvement 


wcampbell0x2a:
Some thoughts on the improvement of performance within `deku`. `read` has been changed to just return the amount of bytes read. So far I only implemented the new `read` for u8 for some initial testing. I haven't messed around with benchmarking or implementing for other types, just wanted to see if anyone had objections before I put some time into that (when I get more time... someday ;) ) 

I'm assuming this could be done for `write` primitives also, but haven't tried that.
```rust
/// "Reader" trait: read bits and construct type
pub trait NewDekuRead<'a, Ctx = ()> {
    /// Read bits and construct type
    /// * **input** - Input as bits
    /// * **ctx** - A context required by context-sensitive reading. A unit type `()` means no context
    /// needed.
    ///
    /// Returns the remaining bits after parsing in addition to Self.
    fn read_new(
        input: &'a BitSlice<u8, Msb0>,
        ctx: Ctx,
    ) -> Result<(usize, Self), DekuError>
    where
        Self: Sized;
}
/// "Reader" trait: implemented on DekuRead struct and enum containers. A `container` is a type which
/// doesn't need any context information.
pub trait NewDekuContainerRead<'a>: NewDekuRead<'a, ()> {
    /// Read bytes and construct type
    /// * **input** - Input given as data and bit offset
    ///
    /// Returns the remaining bytes and bit offset after parsing in addition to Self.
    fn from_bytes(input: (&'a [u8], usize)) -> Result<((usize, usize), Self), DekuError>
    where
        Self: Sized;
}

// specialize u8 for ByteSize
impl NewDekuRead<'_, ()> for u8 {
    fn read_new(
        input: &BitSlice<u8, Msb0>,
        _: (),
    ) -> Result<(usize, Self), DekuError> {
        const MAX_TYPE_BITS: usize = BitSize::of::<u8>().0;
        if input.len() < MAX_TYPE_BITS {
            return Err(DekuError::Incomplete(NeedSize::new(
                MAX_TYPE_BITS,
            )));
        }

        // PANIC: We already check that input.len() < bit_size above, so no panic will happen
        let value = input[..MAX_TYPE_BITS].load::<u8>();
        Ok((MAX_TYPE_BITS, value))
    }
}
```

The generated code:
```rust
#[derive(Debug, DekuRead)]
pub struct Test {
    a: u8,
    b: u8,
}

impl NewDekuRead<'_, ()> for Test {
    fn read_new(
[...]

PR #345, opened 2023-06-15
Improve deku performance

wcampbell0x2a:
Still want to do benchmarks and more testing, but it's done-ish

See #344

### DIFF
--- a/benches/deku.rs
+++ b/benches/deku.rs
@@ -105,13 +105,13 @@ fn criterion_benchmark(c: &mut Criterion) {
     });
 
     let deku_read_vec_input = {
-        let mut v = [0xFFu8; 101].to_vec();
+        let mut v = [0xffu8; 101].to_vec();
         v[0] = 100u8;
         v
     };
     let deku_write_vec_input = DekuVec {
         count: 100,
-        data: vec![0xFF; 100],
+        data: vec![0xff; 100],
     };
     c.bench_function("deku_read_vec", |b| {
         b.iter(|| deku_read_vec(black_box(&deku_read_vec_input)))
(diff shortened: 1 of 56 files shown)


v1gnesh:
Hey @wcampbell0x2a I have a dumb question - 
Since in many cases, parsing is for static content (either sizes known fully at compile time or reasonable upper bounds may be known for dynamic content), would it help if memory for these were pre-allocated before getting into read?

wcampbell0x2a:
> Hey @wcampbell0x2a I have a dumb question - Since in many cases, parsing is for static content (either sizes known fully at compile time or reasonable upper bounds may be known for dynamic content), would it help if memory for these were pre-allocated before getting into read?

You can see some tests for allocation counts here: https://github.com/sharksforarms/deku/blob/master/tests/test_alloc.rs. Many of these allocations are quite dynamic based upon bit alignment when reading.

wcampbell0x2a:
> ## Benchmark for [1b73cc4](https://github.com/sharksforarms/deku/commit/1b73cc40d7faecb7827c0d745fc77d05257e3671)
> Click to view benchmark

This commit showed a %5 performance increase for deku `deku_read_byte` on my machine.

sharksforarms:
Yeah not sure how much I trust the CI benchmarking, I'm not sure it's fully accurate. I'll take a look at this when I have more time, thanks for working on this!

wcampbell0x2a:
This breaks the logging feature. I'll need to fix that.

wcampbell0x2a:
I also need to test the old + new with both having LTO enabled. The performance might be worse with this patch.

PR #355, opened 2023-08-18
Impl writer

wcampbell0x2a:
Implement faster writing for bytes and support for `Write`. At the expense of some performance for only bit-wise writing.

# Tested
- https://github.com/wcampbell0x2a/backhand/pull/524
- https://github.com/wcampbell0x2a/asterix-rs/pull/9
- https://github.com/rsadsb/adsb_deku/pull/212
- https://github.com/wcampbell0x2a/cpio-deku

## Create Issue
- It would be nice to specialize for `&[u8]` and `Vec<u8>` so that results directly to one invocation of `writer.write_bytes(..)` instead of 1 byte per len. This is only possible for `Ctx = ()`?

See https://github.com/wcampbell0x2a/deku-bench/pull/6 (that uses https://github.com/sharksforarms/deku/pull/393)

### DIFF
--- a/src/impls/boxed.rs
+++ b/src/impls/boxed.rs
@@ -52,6 +52,18 @@ where
     }
 }
 
+impl<T, Ctx> DekuWriter<Ctx> for Box<T>
+where
+    T: DekuWriter<Ctx>,
+    Ctx: Copy,
+{
+    /// Write all `T`s to bits
+    fn to_writer<W: Write>(&self, writer: &mut Writer<W>, ctx: Ctx) -> Result<(), DekuError> {
+        self.as_ref().to_writer(writer, ctx)?;
+        Ok(())
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use no_std_io::io::Cursor;
--- a/src/impls/boxed.rs
+++ b/src/impls/boxed.rs
@@ -1,12 +1,12 @@
-use no_std_io::io::Read;
+use no_std_io::io::{Read, Write};
 
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 
-use bitvec::prelude::*;
-
 use crate::ctx::Limit;
-use crate::{DekuError, DekuReader, DekuWrite};
+use crate::reader::Reader;
+use crate::writer::Writer;
+use crate::{DekuError, DekuReader, DekuWriter};
 
 impl<'a, T, Ctx> DekuReader<'a, Ctx> for Box<T>
 where
@@ -14,22 +14,11 @@ where
     Ctx: Copy,
 {
     fn from_reader_with_ctx<R: Read>(
-        reader: &mut crate::reader::Reader<R>,
+        reader: &mut Reader<R>,
         inner_ctx: Ctx,
     ) -> Result<Self, DekuError> {
         let val = <T>::from_reader_with_ctx(reader, inner_ctx)?;
         Ok(Box::new(val))
-    }
-}
-
-impl<T, Ctx> DekuWrite<Ctx> for Box<T>
-where
-    T: DekuWrite<Ctx>,
-    Ctx: Copy,
-{
-    /// Write T from box
-    fn write(&self, output: &mut BitVec<u8, Msb0>, inner_ctx: Ctx) -> Result<(), DekuError> {
-        self.as_ref().write(output, inner_ctx)
     }
 }
 
@@ -40,7 +29,7 @@ where
     Predicate: FnMut(&T) -> bool,
 {
     fn from_reader_with_ctx<R: Read>(
-        reader: &mut crate::reader::Reader<R>,
+        reader: &mut Reader<R>,
         (limit, inner_ctx): (Limit<T, Predicate>, Ctx),
     ) -> Result<Self, DekuError> {
         // use Vec<T>'s implementation and convert to Box<[T]>
@@ -49,15 +38,15 @@ where
     }
 }
 
-impl<T, Ctx> DekuWrite<Ctx> for Box<[T]>
+impl<T, Ctx> DekuWriter<Ctx> for Box<[T]>
 where
-    T: DekuWrite<Ctx>,
+    T: DekuWriter<Ctx>,
     Ctx: Copy,
 {
     /// Write all `T`s to bits
-    fn write(&self, output: &mut BitVec<u8, Msb0>, ctx: Ctx) -> Result<(), DekuError> {
+    fn to_writer<W: Write>(&self, writer: &mut Writer<W>, ctx: Ctx) -> Result<(), DekuError> {
         for v in self.as_ref() {
-            v.write(output, ctx)?;
+            v.to_writer(writer, ctx)?;
         }
         Ok(())
     }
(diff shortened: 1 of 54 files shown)


src/writer.rs
```diff
@@ -0,0 +1,184 @@
...
+#[cfg(feature = "logging")]
+use log;
+
+use crate::DekuError;
+
+const fn bits_of<T>() -> usize {
+    core::mem::size_of::<T>().saturating_mul(<u8>::BITS as usize)
+}
+
+/// Container to use with `from_reader`
```
wcampbell0x2a, line 16: fix copy+paste comment

src/writer.rs
```diff
@@ -0,0 +1,193 @@
...
+            self.leftover
+                .chunks_exact(bits_of::<u8>())
+                .zip(buf.iter_mut())
+                .for_each(|(byte, slot)| {
+                    *slot = byte.load_be();
+                });
+
+            if self.inner.write_all(&buf).is_err() {
+                return Err(DekuError::Write);
+            }
```
wcampbell0x2a, line 141: Update `bits_written` here

wcampbell0x2a:
@sharksforarms this is ready for review!

sharksforarms:
@wcampbell0x2a ack. I may not get to review it until the new year

wcampbell0x2a:
Butterfingers closing the mr 😂

No problem! I also might bike-shed the names back to `read` and `write` instead of `to_writer`. Idk.

src/impls/boxed.rs
```diff
@@ -1,46 +1,35 @@
...
-        reader: &mut crate::reader::Reader<R>,
+        reader: &mut Reader<R>,
         inner_ctx: Ctx,
     ) -> Result<Self, DekuError> {
         let val = <T>::from_reader_with_ctx(reader, inner_ctx)?;
         Ok(Box::new(val))
     }
 }
 
-impl<T, Ctx> DekuWrite<Ctx> for Box<T>
```
sharksforarms, line 25: We'd still want `DekuWriter` for `Box` to support:

```rust
#[derive(DekuWrite)]
struct MyStruct {
    field: Box<u8>,
}
```

Without it I'm getting:
```
the trait bound `Box<u8>: DekuWriter<_>` is not satisfied
```

src/impls/boxed.rs
```diff
@@ -85,9 +75,10 @@ mod tests {
         let res_read = <Box<u16>>::from_reader_with_ctx(&mut reader, ()).unwrap();
         assert_eq!(expected, res_read);
 
-        let mut res_write = bitvec![u8, Msb0;];
-        res_read.write(&mut res_write, ()).unwrap();
-        assert_eq!(input.to_vec(), res_write.into_vec());
+        let mut out_buf = vec![];
+        let mut writer = Writer::new(&mut out_buf);
+        res_read.to_writer(&mut writer, ()).unwrap();
+        assert_eq!(input.to_vec(), out_buf.to_vec());
```
sharksforarms, line 81: nit/optional:Here's an alternative which I slightly prefer (one less variable), there's a couple locations that use this pattern

```suggestion
        let mut writer = Writer::new(vec![]);
        res_read.to_writer(&mut writer, ()).unwrap();
        assert_eq!(input.to_vec(), writer.inner);
```

src/impls/slice.rs
```diff
@@ -95,72 +93,17 @@ mod tests {
...
-            bits![u8, Msb0;],
-        ),
-        case::normal_le(
-            [0xDD, 0xCC, 0xBB, 0xAA, 0x99, 0x88, 0x77, 0x66].as_ref(),
-            Endian::Big,
-            [[0xDDCC, 0xBBAA], [0x9988, 0x7766]],
-            bits![u8, Msb0;],
-        ),
-    )]
-    fn test_nested_array_bit_read(
```
sharksforarms, line 124: curious why these were removed

src/writer.rs
```diff
@@ -0,0 +1,200 @@
...
+        let mut buf = alloc::vec![0x00; bits.len() / 8];
+        let mut count = 0;
+        bits.chunks_exact(bits_of::<u8>())
+            .zip(buf.iter_mut())
+            .for_each(|(byte, slot)| {
+                *slot = byte.load_be();
+                count += 1;
+            });
+
+        // SAFETY: This does noto have a safety comment in bitvec. But this is safe
```
sharksforarms, line 78:
```suggestion
        // SAFETY: This does not have a safety comment in bitvec. But this is safe
```

src/writer.rs
```diff
@@ -0,0 +1,200 @@
...
+        // because of `count` here will always still be within the bounds
+        // of `bits`
+        bits = unsafe { bits.get_unchecked(count * bits_of::<u8>()..) };
+
+        // TODO: with_capacity?
```
sharksforarms, line 83: TODO, clarify or address

Do you mean we could infer `self.leftover` capacity at initialization?

src/writer.rs
```diff
@@ -0,0 +1,200 @@
...
+    #[inline(always)]
+    pub fn write_bytes(&mut self, buf: &[u8]) -> Result<(), DekuError> {
+        #[cfg(feature = "logging")]
+        log::trace!("writing {} bytes: {buf:02x?}", buf.len());
+
+        if !self.leftover.is_empty() {
+            #[cfg(feature = "logging")]
+            log::trace!("leftover exists");
+
+            // TODO: we could check here and only send the required bits to finish the byte?
```
sharksforarms, line 110: Should this TODO be addressed or it's a potential optimization?

src/writer.rs
```diff
@@ -0,0 +1,200 @@
...
+            #[cfg(feature = "logging")]
+            log::trace!("finalized: {} bits leftover", self.leftover.len());
+
+            // add bits to be byte aligned so we can write
+            self.leftover
+                .extend_from_bitslice(&bitvec![u8, Msb0; 0; 8 - self.leftover.len()]);
+            let mut buf = alloc::vec![0x00; self.leftover.len() / 8];
+
+            // write as many leftover to the buffer (as we can, can't write bits just bytes)
+            // TODO: error if bits are leftover? (not bytes aligned)
```
sharksforarms, line 137: is this covered by some test case? I'm curious what it currently looks like when/if we hit this

sharksforarms (commented):
This looks fantastic! I left a couple of minor comments/questions. Very exciting for this to land with the reader changes + performance improvements, thanks for your persistent efforts here

src/writer.rs
wcampbell0x2a: This is actually can't be hit, as we fill out that buffer to the length of a byte above.

src/writer.rs
wcampbell0x2a: Changed to TODO(perf), as yes, it's only an optimization

src/writer.rs
wcampbell0x2a: Changed, not applicable.

src/impls/boxed.rs
wcampbell0x2a: Changed everywhere

src/impls/boxed.rs
wcampbell0x2a: Good catch, should be good now

src/impls/slice.rs
wcampbell0x2a: No reason, I think I missed this in a rebase. Added back.

sharksforarms (approved):
Sorry for the delay, this looks great.

Issue #105, opened 2020-09-29
Read from a std::io `Read` + `Seek` and output to a `Write`


sharksforarms:
Reader functions can take in a `Read` + `Seek` trait impl, write can write to a `Write`

https://github.com/sharksforarms/deku/issues/100

Feature gated under "std" feature as these are std traits.

For nostd, maybe need to wait for: https://github.com/rust-lang/rust/issues/48331

References:
https://doc.rust-lang.org/std/io/trait.Read.html
https://doc.rust-lang.org/std/io/trait.Write.html
https://doc.rust-lang.org/std/io/trait.Seek.html

CJKay:
If anybody's looking for a way to work around this in the meantime, I hacked this up this nightmare:

```rust
fn read_offset<'a, C, T, O>(
    rest: &'a BitSlice<Msb0, u8>,
    input: &'a BitSlice<Msb0, u8>,
    ctx: C,
    offset: O,
) -> Result<(&'a BitSlice<Msb0, u8>, T), DekuError>
where
    C: Copy,
    T: DekuRead<'a, C>,
    O: Into<usize>,
{
    let offset = offset.into();
    let subslice = input.get(offset..).ok_or_else(|| {
        let need = NeedSize::new(rest.len() - offset);

        DekuError::Incomplete(need)
    })?;

    T::read(subslice, ctx)
}

fn read_offset_count<'a, C, T, U, O, N>(
    rest: &'a BitSlice<Msb0, u8>,
    input: &'a BitSlice<Msb0, u8>,
    ctx: C,
    offset: O,
    count: N,
) -> Result<(&'a BitSlice<Msb0, u8>, T), DekuError>
where
    C: Copy,
    T: DekuRead<'a, (Limit<U, fn(&U) -> bool>, C)>,
    U: Copy + DekuRead<'a, C>,
    O: Into<usize>,
    N: Into<usize>,
{
    read_offset(rest, input, (Limit::new_count(count.into()), ctx), offset)
}
```

Example usage:

```
    /// File name hashes.
    #[deku(reader = "read_offset_count(deku::rest, deku::input_bits, endian, ((*file_name_hashes_offset as usize) + 8) * 8, *file_count as usize)")]
    file_name_hashes: Vec<u64>,
```

TheWillard:
I'm curious on how this could be implemented, since this feature would really help in some of my projects. 
[...]

PR #360, opened 2023-09-04
Impl reader+writer seek

wcampbell0x2a:
- Add `Seek` to `Reader` and `Writer`
- Add `as_mut` to `Reader`, and tests
- Add: `seek_rewind`, `seek_from_current`, `seek_from_end`,  `seek_from_start`,

See https://github.com/sharksforarms/deku/issues/105

Currently being used in https://github.com/wcampbell0x2a/cpio-deku

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- Add attributes `seek_from_start`, `seek_from_current`, `seek_from_end`, and `seek_rewind` to control the position of the reader before reading a field ([#360](https://github.com/sharksforarms/deku/pull/360))
 
 ## [0.17.0] - 2024-04-23
 
@@ -21,7 +22,6 @@
      VariantC((u8, u8)),
  }
 ```
-- Add attributes `seek_from_start`, `seek_from_current`, `seek_from_end`, and `seek_rewind` to control the position of the reader before reading a field ([#360](https://github.com/sharksforarms/deku/pull/360))
 
 ### Updated Reader API
 - Changed API of reading to use `io::Read`, bringing massive performance and usability improvements ([#352](https://github.com/sharksforarms/deku/pull/352))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
      VariantC((u8, u8)),
  }
 ```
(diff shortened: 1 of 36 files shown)


vhdirk:
This would be pretty useful in my project. Anything I can assist with to get this merged?

wcampbell0x2a:
> This would be pretty useful in my project. Anything I can assist with to get this merged?

We need https://github.com/no-std-io/no-std-io/pull/5 and a final review by me and @sharksforarms

wcampbell0x2a:
> > This would be pretty useful in my project. Anything I can assist with to get this merged?
> 
> We need [no-std-io/no-std-io#5](https://github.com/no-std-io/no-std-io/pull/5) and a final review by me and @sharksforarms

Decided to just fork that repo and use my own `no-std-core2` for now, since that wasn't getting merged and i've tried contacting them a couple of times.

sharksforarms (approved):
Thanks! This is great. This got me thinking, it would be pretty cool if there was a way to add attributes like this without adding bounds to the generic, such that "seek_" could only be used if `T` is `Seek`, else it would be a compile error. This way, we don't impose more bounds than necessary?

The downside of this PR is that now types are now required to be `Seek` (maybe less of a concern with the reader interface... but I can for-see some folks not liking this, maybe?)

wcampbell0x2a:
> Thanks! This is great. This got me thinking, it would be pretty cool if there was a way to add attributes like this without adding bounds to the generic, such that "seek_" could only be used if `T` is `Seek`, else it would be a compile error. This way, we don't impose more bounds than necessary?
> 
> The downside of this PR is that now types are now required to be `Seek` (maybe less of a concern with the reader interface... but I can for-see some folks not liking this, maybe?)

I'm not sure about any solutions that includes generic bounds, I think you'll always need to have `Seek` as a bounds on `Reader`, unless i'm wrong.

the friends at `binrw` 'solve' this by having a [NoSeek](https://docs.rs/binrw/latest/binrw/io/struct.NoSeek.html)

sharksforarms:
> the friends at `binrw` 'solve' this by having a [NoSeek](https://docs.rs/binrw/latest/binrw/io/struct.NoSeek.html)

Ah! Indeed. It was likely enough of a pain point to include something in the library.

PR #366, opened 2023-10-07
Fix hygiene in gen_internal_field_ident

interruptinuse:
- Change generated ident prefix to `__deku___` (three underscores) so that emitted field reads are immediately obvious in expanded code, although I think two underscores would be valid as well.
- Also add a field to `test_named_struct`.

### DIFF
--- a/deku-derive/src/macros/mod.rs
+++ b/deku-derive/src/macros/mod.rs
@@ -117,7 +117,7 @@ fn gen_internal_field_ident(ident: &TokenStream) -> TokenStream {
     // Concat token: https://github.com/rust-lang/rust/issues/29599
     let span = ident.span();
     let s = ident.to_string();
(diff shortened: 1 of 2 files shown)


wcampbell0x2a (approved):

sharksforarms (approved):

sharksforarms:
FYI @interruptinuse I'm now see'ing `Variable `__deku___size` should have snake_case name, e.g. `__deku_size` [non_snake_case]` in my editor :eyes:

interruptinuse:
Oh, my bad 👀!

My first instinct was to add `#[allow(non_snake_case)]`, however, it might not be sufficient (rust-lang/rust-analyzer#6541). I'm going to go ahead and change the prefix to `__deku_field_`.

PR #380, opened 2023-12-08
read_with_predicate: stop when empty

vhdirk:
When setting a predicate, I have a situation where the bitvec is empty, yet it still attempts to start reading the next element.
It seems not too weird to assume that we should stop reading when the last element completed successfully and there's nothing more to process left.

### DIFF
--- a/src/impls/hashmap.rs
+++ b/src/impls/hashmap.rs
@@ -29,7 +29,7 @@ fn read_hashmap_with_predicate<
     let mut rest = input;
     let mut found_predicate = false;
 
-    while !found_predicate {
+    while !found_predicate && !rest.is_empty(){
         let (new_rest, kv) = <(K, V)>::read(rest, ctx)?;
         found_predicate = predicate(
             unsafe { new_rest.as_bitptr().offset_from(input.as_bitptr()) } as usize,
--- a/src/impls/hashset.rs
+++ b/src/impls/hashset.rs
@@ -28,7 +28,7 @@ fn read_hashset_with_predicate<
     let mut rest = input;
     let mut found_predicate = false;
 
-    while !found_predicate {
+    while !found_predicate && !rest.is_empty() {
         let (new_rest, val) = <T>::read(rest, ctx)?;
         found_predicate = predicate(
             unsafe { new_rest.as_bitptr().offset_from(input.as_bitptr()) } as usize,
--- a/src/impls/slice.rs
+++ b/src/impls/slice.rs
@@ -28,7 +28,7 @@ where
         let read_idx = unsafe { rest.as_bitptr().offset_from(input.as_bitptr()) } as usize;
         value = input[..read_idx].domain().region().unwrap().1;
 
-        if predicate(read_idx, &val) {
+        if predicate(read_idx, &val) || rest.is_empty() {
             break;
         }
     }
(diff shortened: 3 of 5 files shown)


sharksforarms:
Thanks for the PR! I think having this pass instead of error may be confusing as we expect the predicate to finish as true.

In your binary format, is there a `size` or another field that indicates there is data to be read into your Vec? Normally there is something like this to prevent reading arbitrary data.

Here's an example using `cond` 

```
use deku::prelude::*;
use std::convert::TryFrom;

struct TestStruct {
    size: u8,
    #[deku(cond = "*size > 0", until = "|v: &u8| *v == 0xDD")]
    data: Vec<u8>,
}

fn main() {
    env_logger::init();
    {
        let test_data: Vec<u8> = [0x00].to_vec();
        let ret_read = TestStruct::try_from(test_data.as_ref()).unwrap();
        assert_eq!(
            TestStruct {
                size: 0x00,
                data: vec![]
            },
            ret_read
        );
    }
    {
        let test_data: Vec<u8> = [0x03, 0xCC, 0xAA, 0xBB].to_vec();
        let ret_read = TestStruct::try_from(test_data.as_ref()).unwrap();
        assert_eq!(
            TestStruct {
                size: 0x03,
                data: vec![0xCC, 0xAA, 0xBB]
            },
            ret_read
        );
    }
}
```

vhdirk:
You're quite right. 
The protocol I'm implementing does have a size field, so I've refactored my code to reflect that. However, the size field is at the beginning of the frame, whereas the array I'm reading is somewhere further in the message.

Quite rudimentary, it looks like this:

```
| frame_size | control | [T; frame_size - sizeof(control)]  |
```

with the added caveat that the `control` field itself is also not a fixed size, it's size depends on some internal flags.
Do you have an idea from the top of your head how I could tackle this (elegantly) ?

vhdirk:
No, it's somewhat more involved than that:

```
struct TestControl {
   #[deku(bits="1", pad_bits_after="7")]
   flag: bool;

    #[deku(cond = "*flag")]
    value: Option<u16>,
}

struct TestStruct {
    size: u8,
    control: TestControl,
    #[deku(cond = "(*size - control.calculated_packed_size() ) > 0", until = "|v: &u8| *v == 0xDD")]
    data: Vec<u8>,
}
```

In this simple example, I could check `TestControl::flag` again in the `cond` attribute in `TestStruct`, but in the protocol at hand there would simple be too many variables to properly do this. I reckon I'd a placeholder along the lines of `deku::bytes_read` that would propagat the number of bytes read up until that point...

sharksforarms:
> No, it's somewhat more involved than that:
> 
> ```
> #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
> struct TestControl {
>    #[deku(bits="1", pad_bits_after="7")]
>    flag: bool;
> 
>     #[deku(cond = "*flag")]
>     value: Option<u16>,
> }
> 
> #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
> struct TestStruct {
>     size: u8,
>     control: TestControl,
>     #[deku(cond = "(*size - control.calculated_packed_size() ) > 0", until = "|v: &u8| *v == 0xDD")]
>     data: Vec<u8>,
> }
> ```
> 
> In this simple example, I could check `TestControl::flag` again in the `cond` attribute in `TestStruct`, but in the protocol at hand there would simple be too many variables to properly do this. I reckon I'd a placeholder along the lines of `deku::bytes_read` that would propagate the number of bytes read up until that point...

There is `deku::byte_offset` which is what you describe, would that help your use case?

https://docs.rs/deku/latest/deku/#internal-variables-and-previously-read-fields

vhdirk:
Oh, not sure how I missed that. byte_offset is perfect! thanks!

PR #381, opened 2023-12-08
Fix non_snake_case diagnostic in derive::DekuRead

interruptinuse:
Fixes https://github.com/sharksforarms/deku/pull/366#issuecomment-1847412600.


sharksforarms (approved):
ty!

Issue #100, opened 2020-09-21
Pointer: jump to position, parse/write struct, and come back


marnunez:
In construct, there's a wrapper called [Pointer](https://construct.readthedocs.io/en/latest/streaming.html#field-wrappers), which is really useful for binary files that have jumptables. It uses a field as an offset, jumps to the position, parses/writes the content of the structure, and jumps back to position. Would something like this be feasible in Deku?

sharksforarms:
I don't see why not! Possibly a `seek` and `peek` attribute. Do you have an example protocol or format for reference?

marnunez:
There's a nice example of case of use in the construct gallery for the [PE/COFF](https://github.com/construct/construct/blob/master/gallery/pe32coff.py) format.

kitlith:
So, I went to implement something like this as part of my own project, and ran into the fact that currently, an implementation of DekuRead has no way to perform an absolute offset operation -- you'd have to pass in the original BitSlice as context. so, that indicates to me that some breaking changes to deku are likely necessary to make any sort of file pointer ergonomic, or else some very careful additions to the derive macro to thread the starting position through to whatever needs it.

sharksforarms:
@kitlith Breaking changes are fine with me. This crate is still pre 1.0.0 :)

CJKay:
For anybody looking for an interim solution, I hacked up this absolute nightmare:

```rust
fn read_offset<'a, C, T, O>(
    rest: &'a BitSlice<Msb0, u8>,
    input: &'a BitSlice<Msb0, u8>,
    ctx: C,
    offset: O,
) -> Result<(&'a BitSlice<Msb0, u8>, T), DekuError>
where
    C: Copy,
    T: DekuRead<'a, C>,
    O: Into<usize>,
{
    let offset = offset.into();
    let subslice = input.get(offset..).ok_or_else(|| {
        let need = NeedSize::new(rest.len() - offset);

        DekuError::Incomplete(need)
    })?;

    T::read(subslice, ctx)
}

fn read_offset_count<'a, C, T, U, O, N>(
    rest: &'a BitSlice<Msb0, u8>,
    input: &'a BitSlice<Msb0, u8>,
    ctx: C,
    offset: O,
    count: N,
) -> Result<(&'a BitSlice<Msb0, u8>, T), DekuError>
where
    C: Copy,
    T: DekuRead<'a, (Limit<U, fn(&U) -> bool>, C)>,
[...]

PR #387, opened 2023-12-27
Add read_all attribute

### LINKED ISSUE #230: `read_all` attribute
Currently there are the `count`, `bytes_read`, `bits_read` and `until` attributes you can apply to a `Vec` to tell deku how much to read.

A `read_all` (name undecided) attribute that tells deku to parse till the end of the input would be very handy and could be used to avoid some ugly math in the `bytes_read` attribute.

wcampbell0x2a:
Add read_all attribute to read until the reader.end() returns true

Fix couple of doc bugs

Closes #230

### DIFF
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -387,6 +387,9 @@ struct FieldData {
     /// a predicate to decide when to stop reading elements into the container
     until: Option<TokenStream>,
 
+    /// read until `reader.end()`
+    read_all: bool,
+
     /// apply a function to the field after it's read
     map: Option<TokenStream>,
 
@@ -454,6 +457,7 @@ impl FieldData {
             bits_read: receiver.bits_read?,
             bytes_read: receiver.bytes_read?,
             until: receiver.until?,
+            read_all: receiver.read_all,
             map: receiver.map?,
             ctx,
             update: receiver.update?,
@@ -518,6 +522,18 @@ impl FieldData {
             return Err(cerror(
                 data.default.span(),
                 "`default` attribute cannot be used here",
+            ));
+        }
+
+        // Validate usage of read_all
+        if data.read_all
+            && (data.until.is_some()
+                || data.count.is_some()
+                || (data.bits_read.is_some() || data.bytes_read.is_some()))
+        {
+            return Err(cerror(
+                data.bits.span(),
+                "conflicting: `read_all` cannot be used with `until`, count`, `bits_read`, or `bytes_read`",
             ));
         }
 
@@ -758,6 +774,10 @@ struct DekuFieldReceiver {
     /// a predicate to decide when to stop reading elements into the container
     #[darling(default = "default_res_opt", map = "map_litstr_as_tokenstream")]
     until: Result<Option<TokenStream>, ReplacementError>,
+
+    /// read until `reader.end()`
+    #[darling(default)]
+    read_all: bool,
 
     /// apply a function to the field after it's read
     #[darling(default = "default_res_opt", map = "map_litstr_as_tokenstream")]
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -683,6 +683,17 @@ fn emit_field_read(
                     (::#crate_::ctx::Limit::new_until(#field_until), (#read_args))
                 )?
             }
+        } else if f.read_all {
+            quote! {
+                {
+                    use core::borrow::Borrow;
+                    #type_as_deku_read::from_reader_with_ctx
+                    (
+                        __deku_reader,
+                        (::#crate_::ctx::Limit::end(), (#read_args))
+                    )?
+                }
+            }
         } else {
             quote! {
                 #type_as_deku_read::from_reader_with_ctx
--- a/examples/read_all.rs
+++ b/examples/read_all.rs
@@ -1,0 +1,23 @@
+use deku::prelude::*;
+use std::convert::{TryFrom, TryInto};
+
+fn main() {
+    #[derive(PartialEq, Debug, DekuRead, DekuWrite)]
+    struct TestStruct {
+        #[deku(read_all)]
+        data: Vec<(u8, u8)>,
+    }
+
+    let test_data: Vec<u8> = [0xaa, 0xbb, 0xcc, 0xdd].to_vec();
+
+    let ret_read = TestStruct::try_from(test_data.as_slice()).unwrap();
+    assert_eq!(
+        TestStruct {
+            data: vec![(0xaa, 0xbb), (0xcc, 0xdd)]
+        },
+        ret_read
+    );
+
+    let ret_write: Vec<u8> = ret_read.try_into().unwrap();
+    assert_eq!(test_data, ret_write);
+}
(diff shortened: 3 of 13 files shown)


duskmoon314:
Looks great! Thanks for your work.

deku-derive/src/lib.rs
```diff
@@ -521,6 +525,16 @@ impl FieldData {
             ));
         }
 
+        // Validate usage of read_all
+        if data.read_all
+            && (data.count.is_some() || (data.bits_read.is_some() || data.bytes_read.is_some()))
+        {
+            return Err(cerror(
+                data.bits.span(),
+                "conflicting: `read_all` cannot be used with `count`, `bits_read`, or `bytes_read`",
```
sharksforarms, line 534: Should `until` be included here?

wcampbell0x2a: Yes. pushed to reflect changes.

deku-derive/src/lib.rs
```diff
@@ -521,6 +525,18 @@ impl FieldData {
...
 
+        // Validate usage of read_all
+        if data.read_all
+            && (data.until.is_some()
+                || data.count.is_some()
+                || (data.bits_read.is_some() || data.bytes_read.is_some()))
+        {
+            return Err(cerror(
+                data.bits.span(),
+                "conflicting: `read_all` cannot be used with `until`, count`, `bits_read`, or `bytes_read`",
```
passcod, line 536:
```suggestion
                "conflicting: `read_all` cannot be used with `until`, `count`, `bits_read`, or `bytes_read`",
```

passcod:
Sometimes it's useful to have both `read_all` and `until`. Here's a workaround I made for this purpose, based on this branch:

```rust
#[derive(Debug)]
struct VecTilEnd<T>(Vec<T>);
impl<'a, T, Ctx, Predicate> DekuReader<'a, (deku::ctx::Limit<T, Predicate>, Ctx)> for VecTilEnd<T>
where
	T: DekuReader<'a, Ctx>,
	Ctx: Copy,
	Predicate: FnMut(&T) -> bool,
{
	fn from_reader_with_ctx<R: std::io::Read>(
		reader: &mut deku::reader::Reader<R>,
		(limit, inner_ctx): (deku::ctx::Limit<T, Predicate>, Ctx),
	) -> Result<Self, DekuError>
	where
		Self: Sized,
	{
		let deku::ctx::Limit::Until(mut predicate, _) = limit else {
			return Err(DekuError::Parse("`until` is required here".to_string()));
		};

		let mut res = Vec::new();
		let start_read = reader.bits_read;

		loop {
			if reader.end() {
				break;
			}
			let val = <T>::from_reader_with_ctx(reader, inner_ctx)?;
			res.push(val);
			if predicate(res.last().unwrap()) {
				break;
			}
		}

		Ok(Self(res))
	}
}

#[derive(Debug, DekuRead)]
struct MaybeProperty {
	#[deku(until = "|b: &u8| *b != b'\\n'")]
	ws: VecTilEnd<u8>,

	#[deku(cond = "ws.0.last().map_or(false, |c| c != &b'[')")]
	property: Option<Property>,
}
```

<sub>(For this very silly thing: https://gist.github.com/passcod/9148782657f3f40df75ff85b8dd77c2a)</sub>

sharksforarms:
@passcod could you clarify your use case with some examples? Trying to understand what you're trying to achieve. I attempted something here, but I didn't need `VecTilEnd` and could use a `Vec` to get the same result. I suspect I'm missing something with the "til end" portion.

```rust
#[derive(Debug, PartialEq, Eq, DekuRead)]
struct MaybePropertyVecTilEnd {
    #[deku(until = "|b: &u8| *b != b'\\n'")]
    ws: VecTilEnd<u8>,

    field: u8,
}

#[derive(Debug, PartialEq, Eq, DekuRead)]
struct MaybePropertyVec {
    #[deku(until = "|b: &u8| *b != b'\\n'")]
    ws: Vec<u8>,

    field: u8,
}

fn main() {
    {
        let mut input = Cursor::new(b"\n\n\n[a");
        let (_, prop) = MaybePropertyVecTilEnd::from_reader((&mut input, 0)).unwrap();
        assert_eq!(
            prop,
            MaybePropertyVecTilEnd {
                ws: VecTilEnd(b"\n\n\n[".to_vec()),
                field: 0x61,
            }
        );
    }

    {
        let mut input = Cursor::new(b"\n\n\n[a");
        let (_b, prop) = MaybePropertyVec::from_reader((&mut input, 0)).unwrap();
        assert_eq!(
            prop,
            MaybePropertyVec {
                ws: b"\n\n\n[".to_vec(), // same as VecTilEnd?
                field: 0x61,
            }
        );
    }
}
```

passcod:
I've added the input file to the [gist](https://gist.github.com/passcod/9148782657f3f40df75ff85b8dd77c2a#file-python-http-service).

The way I reasoned about it is that I needed to read `ws` (which is a few layers deep in the structure) "until either a match or EOF".

```ini

[Section]
Key=value

```

For example this is:

```rust
"\n[Section]\nKey=value\n"
```

and so I read (in the full example in gist):

```plain
Unit::head_ws:      (read until '['):       \n[
Section::name:      (read until ']'):       Section]
MaybeProperty::ws:  (read until not '\n'):  \nK
Property::key:      (read until '='):       ey=
Property::value:    (read until '\n'):      value\n
MaybeProperty::ws:  (read until not '\n'):  !!! need one more byte !!!
(error!)
```

With `VecTilEnd` it continues instead:

```plain
MaybeProperty::ws:      (read until not '\n' or EOF):  *EOF* -> empty vec
MaybeProperty::property (not read because ws is empty -> None)
Section::properties     (closes out because MaybeProperty::property is None)
Unit::sections          (closes out because Section is done)
(successful parse!)
```

sempervictus:
handy, thank you - works for me.

sharksforarms:
@wcampbell0x2a what do you think of the above?

Could be neat to have an indicator of "eof" which could be used in attributes such as:

```rust
#[deku(until = "|v: &u8| *v == 0 || deku::eof")]
data: Vec<u8>
```

sharksforarms (approved):

passcod:
I was thinking about this and I'm not sure (out of ignorance; haven't looked) if it would work with deku internals, but something like

```
until_next = |b: Option<T>| { ... }
```

that would 'peek' at the next byte/bit/item (where `None` is EOF or 'no more data available') and stop at the current if it returns true. That could cover both the EOF case and some part of #100.

wcampbell0x2a:
> @wcampbell0x2a what do you think of the above?
> 
> Could be neat to have an indicator of "eof" which could be used in attributes such as:
> 
> ```rust
> #[deku(until = "|v: &u8| *v == 0 || deku::eof")]
> data: Vec<u8>
> ```
> 
> I'm ok with adding a `read_all` attribute to facilitate this use case, but thinking about others such as the one mentioned by @\passcod

How would this convert into a `Limit(_)` ?

wcampbell0x2a:
@sharksforarms I added a `Add-To-Changelog` label, which is something I usually use to keep track of issues/MRs that are merged that don't have a changelog entry before releasing. I you don't need/use them lmk

Issue #358, opened 2023-08-27
Experiment with nightly const-trait-impl


wcampbell0x2a:
<details><summary>Previous Issue</summary>
<p>
@wcampbell0x2a it could call all those trait methods at compile time if possible and/or optimize them out:
![image](https://user-images.githubusercontent.com/108802164/235850935-15d0a95c-5733-48ac-8dc2-fb14f52ebbfc.png)
see [feature const-trait-impl](https://doc.rust-lang.org/beta/unstable-book/language-features/const-trait-impl.html)

I'm not sure how `#[inline]` interacts with this. I'm guessing if the call can't be optimized out at compile time, it inlines. It could also be the case that constness is only relevant to the crate where it is defined.
See the last line here (which I used on some `From` impls lately):

> None: The compiler will decide itself if the function should be inlined. This will depend on factors such as the optimization level and the size of the function. Non-generic functions will never be inlined across crate boundaries unless link-time optimization is used; generic functions might be.

> #[inline]: This suggests that the function should be inlined, including across crate boundaries.

Then again, I might have mixed this up (even non-const stuff can get computed at compile time), so if someone wants to go at this, it's probably best to first try to put `#[inline]` on those trait methods and compare asm outputs.
Edit: I quickly tried adding `#[inline]` to all the `read`s and `write`s in `impls::primitives`, it didn't inline (and I wouldn't use inline(always)).

</p>
</details>

_Originally posted by @hecatia-elegua in https://github.com/sharksforarms/deku/issues/308#issuecomment-1532593054_

wcampbell0x2a:
I don't see this as a possible advantage for the time being, and not interesting in maintaining a nightly only feature set.

Issue #308, opened 2023-01-09
Add "bitfield" to Cargo.toml keywords


hecatia-elegua:
This would make it easier to find, I only noticed this existed by reading bitvec docs.
Adding that word in the docs somewhere might help as well, either saying "you can use it like that" or specifying why it is different.

hecatia-elegua:
I might have misunderstood deku's purpose. Is it really meant for any kind of bitfield?

Comparing asm output of some bitfields implemented in deku vs other bitfield crates, deku produces 20x more code.
This is, with `deku = "0.16"` using `cargo-show-asm`.

When looking at `cargo expand`, I noticed the cause is likely `DekuRead` and Co. not being const. It would be very interesting to see what happens if it were (nightly) const behind a feature gate.

could open another issue if you want

wcampbell0x2a:
> When looking at cargo expand, I noticed the cause is likely DekuRead and Co. not being const. It would be very interesting to see what happens if it were (nightly) const behind a feature gate.

Could you explain the nightly feature and what this would do? I'm not familiar  and curious.

hecatia-elegua:
@wcampbell0x2a it could call all those trait methods at compile time if possible and/or optimize them out:
![image](https://user-images.githubusercontent.com/108802164/235850935-15d0a95c-5733-48ac-8dc2-fb14f52ebbfc.png)
see [feature const-trait-impl](https://doc.rust-lang.org/beta/unstable-book/language-features/const-trait-impl.html)

I'm not sure how `#[inline]` interacts with this. I'm guessing if the call can't be optimized out at compile time, it inlines. It could also be the case that constness is only relevant to the crate where it is defined.
See the last line here (which I used on some `From` impls lately):

> None: The compiler will decide itself if the function should be inlined. This will depend on factors such as the optimization level and the size of the function. Non-generic functions will never be inlined across crate boundaries unless link-time optimization is used; generic functions might be.

> #[inline]: This suggests that the function should be inlined, including across crate boundaries.

wcampbell0x2a:
fyi: bitvec needs to use heavy usage of inline already to get decent performance: https://github.com/search?q=repo%3Aferrilab%2Fbitvec+inline&type=code

PR #393, opened 2023-12-27
Improve performance after `impl-writer` and `impl-reader`

wcampbell0x2a:
This fixes *most* performance issues where the rustc compiler in different version with different parameters didn't inline several functions, both generated and `impl`'ed in deku.

See https://github.com/wcampbell0x2a/deku-bench/pull/6 for benchmarks. Speaking of which I will probably add these benchmarks directly into this project at some point.

I actually opened a ticked for `rust-lang`, since the nightly changes regressed the performance of this crate, until this commit was added: https://github.com/rust-lang/rust/issues/118674

This might removed the need for testing: https://github.com/sharksforarms/deku/issues/358, and discussed with https://github.com/sharksforarms/deku/issues/308#issuecomment-1532593054.

See https://github.com/sharksforarms/deku/issues/25

### DIFF
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -117,6 +117,7 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
 
     tokens.extend(quote! {
         impl #imp ::#crate_::DekuReader<#lifetime, #ctx_types> for #ident #wher {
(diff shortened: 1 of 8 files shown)


sharksforarms (dismissed):

wcampbell0x2a:
@sharksforarms rebased, should be good now

sharksforarms (approved):

PR #405, opened 2024-01-17
Add no-assert-string feature

wcampbell0x2a:
When no-assert-string is enabled, remove string generation in .rodata section. When writing parsers made for embedded platforms that are limited in storage these strings add up when many assertions are added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ default = ["std"]
 std = ["deku_derive/std", "bitvec/std", "alloc", "no_std_io/std"]
 alloc = ["bitvec/alloc"]
 logging = ["deku_derive/logging", "log"]
+no-assert-string = ["deku_derive/no-assert-string"]
 
 [dependencies]
 deku_derive = { version = "^0.16.0", path = "deku-derive", default-features = false}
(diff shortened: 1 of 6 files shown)


sempervictus:
Wonder if some sort of rainbow table/heuristic dictionary or compression based approach would help provide middle ground for more usable string space with qualified lookup/decompression speed. Some sort of checksum dictionary stored in runtime might also be more efficient since its not really the bytes of the strings we're usually asserting but their equivalence to X or Y which matches pre-computed hashes.

wcampbell0x2a:
> Wonder if some sort of rainbow table/heuristic dictionary or compression based approach would help provide middle ground for more usable string space with qualified lookup/decompression speed. Some sort of checksum dictionary stored in runtime might also be more efficient since its not really the bytes of the strings we're usually asserting but their equivalence to X or Y which matches pre-computed hashes.

Interesting idea, it could return a hash of both the struct/enum and field which could be reversed at runtime if you wanted that information. That would take some design, this is easier for me to just turn on and not have extra bytes of data..

sharksforarms (approved):

Issue #408, opened 2024-01-20
Context passing (possibly due to multiple elements) creates opaque error


sempervictus:
Using non-adjacent fields to determine and then read enum types per https://github.com/sharksforarms/deku/blob/master/examples/enums.rs appears to get convoluted when endianness and identifiers are in the context. Specifically:
```rust
#[deku_derive(DekuRead, DekuWrite)]
#[derive(Debug, PartialEq)]
#[deku(endian = "endian", ctx = "endian: deku::ctx::Endian")]
pub(super) struct Svc {
    pub(super) hdr: SvcHeader,
    #[deku(ctx = "*hdr")]
    pub(super) svc: SvcBody
}
#[deku_derive(DekuRead, DekuWrite)]
#[derive(Debug, PartialEq)]
#[deku(endian = "endian", ctx = "endian: deku::ctx::Endian")]
pub(super) struct SvcHeader {
    pub(super) svc_type: u16,
    pub(super) pkts: u16,
    #[deku(assert = "*tail == 0")]
    pub(super) tail: u32,
}
#[deku_derive(DekuRead, DekuWrite)]
#[derive(Debug, PartialEq)]
#[deku(endian = "endian", ctx = "endian: deku::ctx::Endian, hdr: SvcHeader", id = "hdr.svc_type")]
pub enum SvcBody {
    #[deku(id = "0x0001")]
    FIRST {
...
```
doesn't compile due to:
```rust
151 | #[deku_derive(DekuRead, DekuWrite)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: message: called `Result::unwrap()` on an `Err` value: Error("expected `,`")
```
being thrown from the `SvcBody` enum

sempervictus:
This is caused, somehow by the act of passing a struct instead of a primitive type in the ctx; passing the `svc_type` as a `u16` resolves the compiler error; but this means having to decompose all structs into members if passing something with complex references.

Concurrently, the enum is eating or missing bytes compared to a vector of 
```rust
#[deku_derive(DekuRead, DekuWrite)]
#[derive(Debug, PartialEq)]
#[deku(endian = "endian", ctx = "endian: deku::ctx::Endian")]
pub(super) struct SvcFirst {
    pub(super) hdr: SvcHeader,
    #[deku(ctx = "*hdr")]
    pub(super) svc: SvcBodyFirst
}
#[deku_derive(DekuRead, DekuWrite)]
[...]

Issue #400, opened 2024-01-06
Revisit enum id & type


sharksforarms:
Context: https://github.com/sharksforarms/deku/pull/398#issuecomment-1875711693

PR #411, opened 2024-01-21
Fix error for invalid deku_id generation on generic enum 

wcampbell0x2a:
- Add/Fixup lifetime for generated code for enum ext id finding. Lifetime __deku needs to be added to the deku_id_type, and the impl.

NOTE: I'm not sure if `__deku` the the best answer, and _now_ ignoring generics is more of a problem of me fixing the problem and the time I have to fix it in a day.

See #408, this also deals with #400


sharksforarms (dismissed):
lgtm, a test case would be good for this too

wcampbell0x2a:
> lgtm, a test case would be good for this too

Added!

sharksforarms (approved):

PR #429, opened 2024-04-20
Add reader::read_bytes_const

wcampbell0x2a:
* Reduce codegen for known sizes at compile time
  - DekuReader<_, Endian> for all types
  - DekuReader<_, (Endian, Endian)> for u8

### DIFF
--- a/src/impls/primitive.rs
+++ b/src/impls/primitive.rs
@@ -79,8 +79,9 @@ impl DekuReader<'_, (Endian, ByteSize)> for u8 {
         reader: &mut Reader<R>,
         (endian, size): (Endian, ByteSize),
     ) -> Result<u8, DekuError> {
-        let mut buf = [0; core::mem::size_of::<u8>()];
-        let ret = reader.read_bytes(size.0, &mut buf)?;
+        const MAX_TYPE_BYTES: usize = core::mem::size_of::<u8>();
+        let mut buf = [0; MAX_TYPE_BYTES];
+        let ret = reader.read_bytes_const::<MAX_TYPE_BYTES>(&mut buf)?;
         let a = match ret {
             ReaderRet::Bytes => <u8>::from_be_bytes(buf),
             ReaderRet::Bits(bits) => {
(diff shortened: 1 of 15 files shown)


src/reader.rs
```diff
@@ -232,6 +233,39 @@ impl<'a, R: Read> Reader<'a, R> {
...
+    ) -> Result<ReaderRet, DekuError> {
+        #[cfg(feature = "logging")]
+        log::trace!("read_bytes: requesting {N} bytes");
+        if self.leftover.is_empty() {
+            if let Err(e) = self.inner.read_exact(buf) {
+                if e.kind() == ErrorKind::UnexpectedEof {
+                    return Err(DekuError::Incomplete(NeedSize::new(N * 8)));
+                }
+
+                // TODO: other errors?
```
sharksforarms, line 256: This `read_exact` is off of the `impl std::io::Read` right? So in theory we could get whatever error from the reader (socket error, file io error, etc.)

Is there a reason we don't want to bubble it up?

wcampbell0x2a: Good idea, opened https://github.com/sharksforarms/deku/pull/433 to address this.

sharksforarms: ty, we can now handle this TODO

sharksforarms (dismissed):
Cool!

wcampbell0x2a:
@sharksforarms should be ready now

src/reader.rs
sharksforarms, line 256: @wcampbell0x2a we could bubble up IO errors here?

wcampbell0x2a: Yep! I won't be back at my computer till later.

wcampbell0x2a: Ready!

src/impls/primitive.rs
```diff
@@ -79,13 +78,15 @@ impl DekuReader<'_, (Endian, ByteSize)> for u8 {
...
-        let ret = reader.read_bytes(size.0, &mut buf)?;
+        const MAX_TYPE_BYTES: usize = core::mem::size_of::<u8>();
+        let mut buf = [0; MAX_TYPE_BYTES];
+        let ret = reader.read_bytes_const::<MAX_TYPE_BYTES>(&mut buf)?;
         let a = match ret {
             ReaderRet::Bytes => <u8>::from_be_bytes(buf),
             ReaderRet::Bits(bits) => {
                 let Some(bits) = bits else {
-                    return Err(DekuError::Parse("no bits read from reader".to_string()));
+                    // todo!();
```
sharksforarms, line 88: nit: comment

sharksforarms (dismissed):
lgtm, just a nit on the remaining todo comment.

src/impls/primitive.rs
wcampbell0x2a: updated

sharksforarms (approved):

PR #434, opened 2024-04-30
Change Strings in DekuError to Cow<'static, str>

wcampbell0x2a:
* Change all strings in DekuError to Cow<'static, str> so that more can be defined as "not alloc'ed strings" and be removed with panic=abort, build-std and friends.

### DIFF
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -1,10 +1,11 @@
 /*!
 Procedural macros that implement `DekuRead` and `DekuWrite` traits
  */
-
 #![warn(missing_docs)]
 
-use std::borrow::Cow;
+extern crate alloc;
+
+use alloc::borrow::Cow;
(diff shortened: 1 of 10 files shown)


sharksforarms:
May be worth at this point to add in the docs about how to reduce binary size, given that there's now a feature + optimizations around it. Could you write a blurb for this in the lib.rs?

wcampbell0x2a:
> May be worth at this point to add in the docs about how to reduce binary size, given that there's now a feature flag + optimizations around it. Could you write a blurb for this in the lib.rs?

Sure, I'll add that tonight

wcampbell0x2a:
> > May be worth at this point to add in the docs about how to reduce binary size, given that there's now a feature flag + optimizations around it. Could you write a blurb for this in the lib.rs?
> 
> Sure, I'll add that tonight

@sharksforarms added!

PR #438, opened 2024-05-14
Update deps

sharksforarms:
Update all dependencies

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,19 +29,19 @@ no-assert-string = ["deku_derive/no-assert-string"]
 [dependencies]
 deku_derive = { version = "^0.16.0", path = "deku-derive", default-features = false}
 bitvec = { version = "1.0.1", default-features = false }
-log = { version = "0.4.17", optional = true }
-no_std_io = { version = "0.5.0", default-features = false, features = ["alloc"] }
-rustversion = "1.0.15"
+log = { version = "0.4.21", optional = true }
+no_std_io = { version = "0.6.0", default-features = false, features = ["alloc"] }
+rustversion = "1.0.16"
 
 [dev-dependencies]
-rstest = "0.18.0"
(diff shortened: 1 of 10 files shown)


wcampbell0x2a:
@sharksforarms looks like some failure.

Also, I think you need to add a codecov ci token to fix the coverage step.

```
error: failed to select a version for `env_logger`.
    ... required by package `deku v0.16.0 (/home/runner/work/deku/deku)`
versions that meet the requirements `^0.11.3` are: 0.11.3

the package `deku` depends on `env_logger`, with features: `anstream` but `env_logger` does not have these features.
 It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
```

sharksforarms:
@wcampbell0x2a updated the msrv to 1.71 (same as env_logger), do you see this being an issue?

I see `1.75.0` in my apt repos, I suspect 1.71 is ok.

Cargo.toml
```diff
@@ -9,7 +9,7 @@ keywords = ["deku", "bits", "serialization", "deserialization", "struct"]
 categories = ["encoding", "parsing", "no-std"]
 description = "bit level serialization/deserialization proc-macro for structs"
 readme = "README.md"
-rust-version = "1.67.1"
+rust-version = "1.71"
```
wcampbell0x2a, line 12: This is also noted in the README. And I would expect a note in the CHANGELOG.

wcampbell0x2a (changes requested):

Cargo.toml
sharksforarms: done in https://github.com/sharksforarms/deku/pull/438/commits/c2e2de945949bd54d6d9e06f62984195de8df162

wcampbell0x2a (approved):

PR #440, opened 2024-05-19
Missing magic changed to be a unique error vs a parsing error

Nub:
This change is needed so parsers that work on stream data can know wether to advance into the buffer to find a valid message without having to manually search for magic bytes before feeding the DekuRead object.

### DIFF
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -444,7 +444,8 @@ fn emit_magic_read(input: &DekuData) -> TokenStream {
                 if *__deku_byte != __deku_read_byte {
                     extern crate alloc;
                     use alloc::borrow::Cow;
-                    return Err(::#crate_::DekuError::Parse(Cow::from(format!("Missing magic value {:?}", #magic))));
+                    return Err(::#crate_::DekuError::Framing(::#crate_::error::NeedMagic::new(&__deku_magic[0..])));
+                    // return Err(::#crate_::DekuError::Parse(Cow::from(format!("Missing magic value {:?}", #magic))));
                 }
             }
         }
--- a/src/error.rs
+++ b/src/error.rs
@@ -33,12 +33,29 @@ impl NeedSize {
     }
 }
 
+/// Magic needed to frame message
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct NeedMagic {
+    magic: Vec<u8>,
+}
+
+impl NeedMagic where {
+    /// Create new [NeedSize] from bits
+    #[inline]
+    pub fn new(magic: &[u8]) -> Self {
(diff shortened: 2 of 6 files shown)


src/error.rs
```diff
@@ -33,12 +33,29 @@ impl NeedSize {
...
 }
 
+/// Magic needed to frame message
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct NeedMagic {
+    magic: Vec<u8>,
+}
+
+impl NeedMagic where {
+    /// Create new [NeedSize] from bits
```
wcampbell0x2a, line 43:
```suggestion
    /// Create new [NeedMagic] from bits
```

deku-derive/src/macros/deku_read.rs
```diff
@@ -444,7 +444,8 @@ fn emit_magic_read(input: &DekuData) -> TokenStream {
                 if *__deku_byte != __deku_read_byte {
                     extern crate alloc;
                     use alloc::borrow::Cow;
-                    return Err(::#crate_::DekuError::Parse(Cow::from(format!("Missing magic value {:?}", #magic))));
+                    return Err(::#crate_::DekuError::Framing(::#crate_::error::NeedMagic::new(&__deku_magic[0..])));
+                    // return Err(::#crate_::DekuError::Parse(Cow::from(format!("Missing magic value {:?}", #magic))));
```
wcampbell0x2a, line 448: remove commented out line

wcampbell0x2a (changes requested):
Thanks! Check the CI, there are some other errors.

Nub:
> Thanks! Check the CI, there are some other errors.

Will do! Shoulda marked this as a draft while I was working on that, apologies.

Nub:
What version of the toolchain is being used, compiler outputs being used for tests without pinning a version is what's causing some of these CI failures.

wcampbell0x2a:
> What version of the toolchain is being used, compiler outputs being used for tests without pinning a version is what's causing some of these CI failures. I had been on a nightly 1.80 which apparently changes the output of the compiler, fun fun!

Yah I usually use `nightly` and then use `cargo +stable test` before commits. The toolchain used in CI is the stable one.

src/error.rs
```diff
@@ -33,12 +33,29 @@ impl NeedSize {
...
+}
+
 /// Deku errors
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum DekuError {
     /// Parsing error when reading
     Incomplete(NeedSize),
+    /// Failed to find magic
+    Framing(NeedMagic),
```
sharksforarms, line 58: would prefer the name relate to magic, unless we plan on expanding this to be an enum with different variants

PR #442, opened 2024-05-22
DRAFT: Add DekuWriterMut

wcampbell0x2a:
Example that this allows. Basically, when you write, you can read from a `Reader`!
```rs
pub enum Data {
    /// On read: Save current stream_position() as `Offset`, seek `header.filesize`
    /// This will be used to seek this this position if we want to extract *just* this file
    Offset(u64),
    /// On write: Write `Reader` to write buffer
    Reader(Box<dyn ReadSeek>),
}

impl DekuReader<'_, u32> for Data {
    fn from_reader_with_ctx<R: Read + Seek>(
        reader: &mut Reader<R>,
        filesize: u32,
    ) -> Result<Data, DekuError> {
        let reader = reader.as_mut();

        // Save the current offset, this is where the file exists for reading later
        let current_pos = reader.stream_position().unwrap();

        // Seek past that file
        let position = filesize as i64 + pad_to_4(filesize as usize) as i64;
        let _ = reader.seek(SeekFrom::Current(position));

        Ok(Self::Offset(current_pos))
    }
}

impl DekuWriterMut for Data {
    fn to_writer_mut<W: Write + Seek>(
        &mut self,
        writer: &mut Writer<W>,
        _: (),
    ) -> Result<(), DekuError> {
        if let Self::Reader(reader) = self {
            // read from reader
            let mut data = vec![];
            reader.read_to_end(&mut data).unwrap();

            // write to deku
            data.to_writer(writer, ())?;

            // add padding
            for _ in 0..pad_to_4(data.len()) {
                0_u8.to_writer(writer, ())?;
            }
        } else {
            panic!("ah");
        }

        Ok(())
    }
}

#[derive(DekuWriteMut)]
pub struct Object {
    // other fields
    data: Data,
}
```

* Add DekuWriterMut, to include parsers that mutate from &mut self during parsing.
* Impls:
  * bool
  * Box
  * IpAddr,
  * NonZero
  * Primitive
  * Slice
  * Vec
  * ()
* Add small example for a Writer that needs Mut

# TODO
- [ ] Make sure the code coverage is good
- [ ] Support Cow?
- [ ] Support HashMap?
- [ ] Support HashSet?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
      VariantC((u8, u8)),
  }
 ```
+- Add attributes `seek_from_start`, `seek_from_current`, `seek_from_end`, and `seek_rewind` to control the position of the reader before reading a field ([#360](https://github.com/sharksforarms/deku/pull/360))
 
 ### Updated Reader API
 - Changed API of reading to use `io::Read`, bringing massive performance and usability improvements ([#352](https://github.com/sharksforarms/deku/pull/352))
--- a/benches/deku.rs
+++ b/benches/deku.rs
@@ -1,4 +1,4 @@
-use std::io::{Cursor, Read};
+use std::io::{Cursor, Read, Seek};
 
 use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion};
 use deku::prelude::*;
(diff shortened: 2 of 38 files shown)


wcampbell0x2a:
I'm not interesting in continuing this MR, I don't even think it's really a good idea lol

PR #446, opened 2024-05-31
Add bits as an optional feature for bitvec

wcampbell0x2a:
This is like the final boss of all my performance work. I try to explain the "no warranty and testing are given for the disable of this feature", while still implementing enough to make this useful in the case of "I want perf to matter".

image

bench: https://github.com/wcampbell0x2a/deku-bench/pull/24

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,6 +22,8 @@ jobs:
       - run: cargo test --all
       # run examples
       - run: cargo run --example 2>&1 | grep -P '   ' | awk '{print $1}' | xargs -i cargo run --example {}
+      # test with no bits feature (don't test docs)
+      - run: cargo test --lib --examples --tests --features std --no-default-features
 
   # Only build on MSRV, since trybuild will fail on older version
   build-msrv:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,3 +56,31 @@ workspace = true
 # Triggers in macro generated code of darling
 # https://github.com/rust-lang/rust-clippy/issues/12643
 manual-unwrap-or-default = "allow"
+
+[[example]]
+name = "custom_reader_and_writer"
+required-features = ["bits"]
+
+[[example]]
+name = "deku_input"
+
+[[example]]
+name = "enums_catch_all"
+required-features = ["bits"]
+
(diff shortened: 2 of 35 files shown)


sharksforarms:
I think the warning is fair, although, what do we need to do to have more assurance and remove that warning?

wcampbell0x2a:
> I think the warning is fair, although, what needs to be done to have more assurance and remove that warning?

I updated this code to behave with running `cargo test --lib --examples --tests --features std --no-default-features`, and added it to CI. This way I am more confident that this options works.

.github/workflows/main.yml
```diff
@@ -22,6 +22,8 @@ jobs:
       - run: cargo test --all
       # run examples
       - run: cargo run --example 2>&1 | grep -P '   ' | awk '{print $1}' | xargs -i cargo run --example {}
+      # test with no bits feature (don't test docs)
```
sharksforarms, line 25: nice

sharksforarms (approved):
very nice, thank you

.github/workflows/pull_request.yml
```diff
@@ -15,3 +15,9 @@ jobs:
         with:
           branchName: ${{ github.base_ref }}
           token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: boa-dev/criterion-compare-action@v3
+        with:
+          branchName: ${{ github.base_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          defaultFeatures: false
```
wcampbell0x2a, line 22: This option doesn't seem to be adding ``--no-default-features`` like it should :/


idk javascript: https://github.com/boa-dev/criterion-compare-action/blob/59f4d964c5f19d7f13d36f5c0944b18ce6652cb0/main.js#L37

wcampbell0x2a: Maybe I fixed it, i'll try out my test version here...

https://github.com/boa-dev/criterion-compare-action/pull/117

wcampbell0x2a: Didn't fix it, guess I won't become a js dev anytime soon 😆

sharksforarms (approved):

PR #452, opened 2024-06-15
Draft: Support arbitrary enum discriminant

### LINKED ISSUE #306: Support `arbitrary_enum_discriminant`
[As of Rust 1.66](https://github.com/rust-lang/rust/releases/tag/1.66.0) the following syntax is added that deku doesn't support yet.

```rust
#[repr(u8)]
enum Foo {
    A(u8) = 0,
    B(i8) = 1,
    C(bool) = 42,
}
```

wcampbell0x2a:
* Add support for arbitrary enum discriminant
* Remove support for defining both `id = 1` and having a discriminant
* Made DekuRead/DekuWrite Id const if possible

Closes #306

### DIFF
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -181,8 +181,15 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
             .first()
             .and_then(|v| v.ident.as_ref())
             .is_some();
+        let disc = &variant.discriminant;
 
         let (use_id, variant_id) = if let Some(variant_id) = &variant.id {
+            if disc.is_some() {
+                return Err(syn::Error::new(
+                    variant.ident.span(),
+                    "DekuRead: `id` cannot be used with arbitrary_enum_discriminant",
+                ));
+            }
             match variant_id {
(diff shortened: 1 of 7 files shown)


sharksforarms:
> Remove support for defining both id = 1 and having a discriminant

I'm not convinced yet we want to do this as it removes some flexibility (what if I want to read/write some arcane binary format, but use it differently/idiomatically in Rust). One example would be identifiers like `0xAB + 1, 0xAB + 2, ...` but in Rust I'd like to represent these as `1, 2, ...`

Alternatively, maybe it's best to say "deku will use the attribute `id` if specified, else it will use the discriminant` wdyt?

wcampbell0x2a:
> > Remove support for defining both id = 1 and having a discriminant
> 
> I'm not convinced yet we want to do this as it removes some flexibility (what if I want to read/write some arcane binary format, but use it differently/idiomatically in Rust). One example would be identifiers like `0xAB + 1, 0xAB + 2, ...` but in Rust I'd like to represent these as `1, 2, ...`
> 
> Alternatively, maybe it's best to say "deku will use the attribute `id` if specified, else it will use the discriminant` wdyt?

Agreed, when I have time to return to this I'll implement that.

PR #454, opened 2024-06-19
Restore v0.16.0 id_pat

wcampbell0x2a:
* When moving to the Read powered impl, I forced the user to use the same type as the id when reading into an id_pat With the use of Seek, we can now fix that and just seek back to the position before the id was read in the case of id_pat

### DIFF
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -227,21 +227,22 @@ fn emit_enum(input: &DekuData) -> Result<TokenStream, syn::Error> {
             .and_then(|v| v.ident.as_ref())
             .is_some();
 
-        let (use_id, variant_id) = if let Some(variant_id) = &variant.id {
+        let mut restore = false;
+        let variant_id = if let Some(variant_id) = &variant.id {
             match variant_id {
-                Id::TokenStream(v) => (false, quote! {&#v}.into_token_stream()),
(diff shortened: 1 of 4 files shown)


deku-derive/src/macros/deku_read.rs
```diff
@@ -589,7 +590,7 @@ fn emit_field_read(
     i: usize,
     f: &FieldData,
     ident: &TokenStream,
-    use_id: bool,
+    restore_pad_id: bool,
```
sharksforarms, line 593: ?
```suggestion
    restore_pat_id: bool,
```

sharksforarms (dismissed):

deku-derive/src/macros/deku_read.rs
```diff
@@ -726,9 +727,18 @@ fn emit_field_read(
             quote!(<#field_type as ::#crate_::DekuReader<'_, _>>)
         };
 
-        if use_id {
+        if restore_pad_id {
```
sharksforarms, line 730: *restore_pat_id

wcampbell0x2a:
Waiting for #360

sharksforarms (dismissed):

wcampbell0x2a:
Ah, I'll need to fix the commits. Unfortunate side effect of squashing.

sharksforarms (approved):

PR #461, opened 2024-07-16
Add to_slice

wcampbell0x2a:


src/lib.rs
```diff
@@ -515,6 +515,32 @@ pub trait DekuContainerWrite: DekuWriter<()> {
...
+    /// let mut buf = [0; 7];
+    /// let s = S { data_00: 0x01, data_01: 0x02, data_02: 0x03 };
+    /// s.to_slice(&mut buf).unwrap();
+    /// assert_eq!(buf, [0x01, 0x02, 0x00, 0x03, 0x00, 0x00, 0x00]);
+    /// ````
+    #[inline(always)]
+    fn to_slice(&self, buf: &mut [u8]) -> Result<(), DekuError> {
+        let mut __deku_writer = Writer::new(buf);
+        DekuWriter::to_writer(self, &mut __deku_writer, ()).unwrap();
+        __deku_writer.finalize().unwrap();
```
amin-mir, line 539: Is it okay to use `unwrap()` here? `to_bytes` is using `?` to propagate the errors to the caller.

wcampbell0x2a: Changed from `unwrap`

src/lib.rs
```diff
@@ -515,6 +515,33 @@ pub trait DekuContainerWrite: DekuWriter<()> {
...
+    /// assert_eq!(buf, [0x01, 0x02, 0x00, 0x03, 0x00, 0x00, 0x00]);
+    /// assert_eq!(amt_written, 7)
+    /// ````
+    #[inline(always)]
+    fn to_slice(&self, buf: &mut [u8]) -> Result<usize, DekuError> {
+        let mut __deku_writer = Writer::new(buf);
+        DekuWriter::to_writer(self, &mut __deku_writer, ())?;
+        __deku_writer.finalize()?;
+
+        Ok(__deku_writer.bits_written / 8)
```
amin-mir, line 542: There might be a corner case here. Would it handle the following case?

```Rust
#[deku(endian = "big")]
struct DekuTest {
    #[deku(bits = 4)]
    field_a: u8,
    field_b: u8,
}
```

Let's say serializing this struct results in bits_written = 12 which means we need two bytes to store them into. But bits_written / 8 = 1. So I'm thinking either `(bit_written + 8) / 8` or `ceil()` would be right way. What do you think?

wcampbell0x2a: The above `finalize` call will fill in bits until it hits the byte boundary to make sure everything is written. This follows the logic of `to_bytes`.

sharksforarms (approved):
Thanks!

PR #464, opened 2024-08-02
Add core_in_error support

stgraham2000:
Add the `core_in_error` feature and implement `core::error::Error` for `DekuError` if the `core_in_error` feature is enabled.

This feature is automatically enabled in Rust nightly so when using Rust nightly you get this feature by default. When using stable Rust, this feature can be optionally added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 authors = ["sharks <sharks@sharks>"]
 edition = "2021"
 name = "deku"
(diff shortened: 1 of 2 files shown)


wcampbell0x2a:
Thanks for the MR!

sharksforarms:
Thanks!

(looks like CI is slightly broken on clippy + others, but tests pass and it doesn't seem related to this PR.)

PR #475, opened 2024-09-04
Qualify types in derive macros

colinjneville:
`derive(DekuRead)` and `derive(DekuWrite)` currently don't work without importing `deku::prelude` because of some unqualified types and trait methods.  
I updated everything unqualified I found in the macros and added a (non-exhaustive) test to ensure they compile without the prelude.


sharksforarms (approved):
wonderful, thank you!

sharksforarms (approved):

Issue #462, opened 2024-07-31
Is there a faster way to read large Vec<u8>?


MeguminSama:
At the moment, say I have a struct like this:

```rs
#[derive(Debug, DekuRead, DekuWrite)]
#[deku(ctx = "size: usize", ctx_default = "0")]
pub struct Rfc {
	#[deku(bytes_read = "size")]
	pub data: Vec<u8>,
}
```

The problem, is that deku seems to loop through the reader for each u8 in bytes_read. This causes it to be ***very*** slow on large vectors. `#[deku(read_all)]` and `#[deku(count = "size")]` are also very slow.

At the moment, we're using our own read function to do something like this:

```rs
match reader.read_bytes(size, &mut buf) {
	Ok(ReaderRet::Bytes) => Ok(Rfc { data: buf }),
	_ => {...}
}
```

Which is significantly faster.

But I was wondering if there was a built-in way to do this with deku, instead of deku looping over each u8? 

If this isn't a feature currently, I might consider implementing it if it's something you'd want in deku.

Thanks!

wcampbell0x2a:
For `read_all` performance, check out the following MR. https://github.com/sharksforarms/deku/pull/441

Since deku makes small repeated reads, using a https://doc.rust-lang.org/std/io/struct.BufReader.html should reduce the read overhead.

MeguminSama:
Thanks for getting back so quickly!

At the moment, our reader is already using a BufReader. We tried doing this in an attempt to speed it up, but unfortunately it's still much too slow when reading the vectors compared to our own read function.

Would some kind of `#[deku(read_buffer = "size")]` attribute be something you'd consider? Or is this out of scope for deku?

wcampbell0x2a:
Definitely try out the merge request, it's really slow without that.

> Would some kind of #[deku(read_buffer = "size")] attribute be something you'd consider? Or is this out of scope for deku?

Sure, I don't have the code in front of me, but I think we only store leftover as a `u8`, so we would need to store the leftovers in a `Vec<u8> ` if needed. I'd like to use that only if you use `read_buffer`, since for embedded platforms you don't want allocations all the time.

wcampbell0x2a:
I also don't know, it could be an improvement in our impl of `Vec`, I think it reads and evaluates one at a time currently.

MeguminSama:
I will take a look, thanks :)

wcampbell0x2a:
[...]

PR #481, opened 2024-09-18
Add count Vec<u8> Specializations

wcampbell0x2a:
* For Vec<u8> when using count, specialize into reading the bytes all
at once

See #462

### DIFF
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -8,6 +8,7 @@ extern crate alloc;
 use alloc::borrow::Cow;
 use std::convert::TryFrom;
 use std::fmt::Display;
+use syn::Type;
 
 use darling::{ast, FromDeriveInput, FromField, FromMeta, FromVariant, ToTokens};
 use proc_macro2::TokenStream;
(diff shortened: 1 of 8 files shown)


src/attributes.rs
```diff
@@ -47,6 +47,7 @@ enum DekuEnum {
 | [bytes_read](#bytes_read) | field | Set the field representing the number of bytes to read into a container
 | [until](#until) | field | Set a predicate returning when to stop reading elements into a container
 | [read_all](#read_all) | field | Read until [reader.end()] returns `true`
+| [read_exact](#read_exact) | field | Read exactly this many bytes into a `Vec<u8>`
```
sharksforarms, line 50: maybe we could have a separate table/section for specializations

deku-derive/src/lib.rs
```diff
@@ -431,6 +432,9 @@ struct FieldData {
     /// tokens providing the number of bytes for the length of the container
     bytes_read: Option<TokenStream>,
 
+    /// read exactly that many bytes
+    read_exact: Option<TokenStream>,
```
sharksforarms, line 436: thinking about this now, could we not specialize `count` for `Vec<u8>` ?

wcampbell0x2a: Yah, proc-macro specializations aren't my favorite, but we don't really have other options... Updated all the code

sharksforarms (dismissed):
Nice!

sharksforarms (approved):
Thanks!

PR #487, opened 2024-10-02
Add NoSeek

wcampbell0x2a:
* Add NoSeek to implement Seek for types that won't need to seek, but have the trait impl.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -377,6 +377,7 @@ pub mod attributes;
 pub mod ctx;
 pub mod error;
 mod impls;
+pub mod noseek;
 pub mod prelude;
 pub mod reader;
 pub mod writer;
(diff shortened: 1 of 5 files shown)


sharksforarms (dismissed):
Thanks! For discovery purposes, it may be nice to have some addition to the docs.

(I'm not sure if a compile 'error'/hint helping the user is possible, but that'd be cool)

wcampbell0x2a:
> Thanks! For discovery purposes, it may be nice to have some addition to the docs.

Sure, will do.

> 
> (I'm not sure if a compile 'error'/hint helping the user is possible, but that'd be cool)


I wish, but I think since 'Seek` isn't a part of this crate, I dont control the diagnostics.

wcampbell0x2a:
@sharksforarms added!

PR #497, opened 2024-10-31
Add ByteSize impl for CString

wcampbell0x2a:
* Use the new dynamic bytes attribute to apply to CString, which in many protocols you already have the size of the CString, so might as well use it

### DIFF
--- a/src/impls/cstring.rs
+++ b/src/impls/cstring.rs
@@ -92,6 +92,14 @@ mod tests {
             &[b'a'],
         ),
 
+        #[should_panic(expected = "Parse(\"Failed to convert Vec to CString: data provided is not nul terminated\")")]
+        case(
+            &[b't', b'e', b's', b't'],
+            Some(4),
+            CString::new("test").unwrap(),
+            &[b'a'],
+        ),
+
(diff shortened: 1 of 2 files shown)


src/impls/cstring.rs
```diff
@@ -49,25 +66,51 @@ mod tests {
...
+            CString::new("test").unwrap(),
+            &[b'a'],
+        ),
+
+        #[should_panic(expected = "Incomplete(NeedSize { bits: 8 })")]
+        case(&[b't', b'e', b's', b't'], Some(5), CString::new("test").unwrap(), &[]),
 
         #[should_panic(expected = "Incomplete(NeedSize { bits: 8 })")]
-        case(&[b't', b'e', b's', b't'], CString::new("test").unwrap(), &[]),
+        case(&[b't', b'e', b's', b't'], None, CString::new("test").unwrap(), &[]),
```
sharksforarms, line 99: thought of a new test, what happens when no null byte is present

and, what happens when it's the correct len, but no nul byte present

wcampbell0x2a: Both of these are represented by 141cc5b5ecd68ba63b9d7b97dbee06964ef2349e

sharksforarms (approved):

PR #502, opened 2024-11-02
Seek first bytes lengths before seek bits for __deku_input

### LINKED ISSUE #500: Offset in from_bytes() causes panic in reader.rs if the bytes are greater than 128
At some point after 0.16 my code started crashing when I was using an offset in from_bytes() where the input bits total more than 128 bytes. So if I have something like this, for example:

```rust
 match AccuracyDescriptionRecord::from_bytes((&data, 1032)) {
    Ok((_remaining_data, acc_desc_record)) => acc_desc_record.print(),
    Err(error) => panic!("Error getting Accuracy Description Record from data: {error}"),
}
``` 

The error looks like:

> thread 'main' panicked at /Users/justin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/deku-0.18.1/src/reader.rs:258:63:
> range end index 434 out of range for slice of length 128

This seems to be happening because from_bytes() is limiting the input offset to the MAX_BITS_AMT in the reader.

wcampbell0x2a:
* For bits input into from_bytes, first seek the byte amount until we can seek the bits amount

Closes #500

### DIFF
--- a/deku-derive/src/macros/deku_read.rs
+++ b/deku-derive/src/macros/deku_read.rs
@@ -42,7 +42,8 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
             {
                 use ::#crate_::no_std_io::Seek;
                 use ::#crate_::no_std_io::SeekFrom;
-                if let Err(e) = __deku_reader.seek(SeekFrom::Current(i64::try_from(#num).unwrap())) {
+                let seek_amt = i64::try_from(#num).expect("could not convert into i64");
+                if let Err(e) = __deku_reader.seek(SeekFrom::Current(seek_amt)) {
                     return Err(::#crate_::DekuError::Io(e.kind()));
                 }
             }
@@ -52,7 +53,8 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
             {
                 use ::#crate_::no_std_io::Seek;
                 use ::#crate_::no_std_io::SeekFrom;
-                if let Err(e) = __deku_reader.seek(SeekFrom::End(i64::try_from(#num).unwrap())) {
+                let seek_amt = i64::try_from(#num).expect("could not convert into i64");
+                if let Err(e) = __deku_reader.seek(SeekFrom::End(seek_amt)) {
                     return Err(::#crate_::DekuError::Io(e.kind()));
                 }
             }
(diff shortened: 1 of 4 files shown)


deku-derive/src/macros/deku_read.rs
```diff
@@ -107,9 +107,19 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
...
             use ::#crate_::DekuReader as _;
             let __deku_reader = &mut deku::reader::Reader::new(__deku_input.0);
-            if __deku_input.1 != 0 {
-                __deku_reader.skip_bits(__deku_input.1)?;
+            let bytes = __deku_input.1 / 8;
+            let bits = __deku_input.1 % 8;
+            if bytes != 0 {
+                use ::#crate_::no_std_io::Seek;
+                use ::#crate_::no_std_io::SeekFrom;
+                if let Err(e) = __deku_reader.seek(SeekFrom::Current(i64::try_from(bytes).unwrap())) {
```
sharksforarms, line 115: should we handle the error from `i64::try_from`? else we should `expect` instead of `unwrap`

deku-derive/src/macros/deku_read.rs
```diff
@@ -121,9 +131,19 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
...
             let mut __deku_cursor = #crate_::no_std_io::Cursor::new(__deku_input.0);
             let mut __deku_reader = &mut deku::reader::Reader::new(&mut __deku_cursor);
-            if __deku_input.1 != 0 {
-                __deku_reader.skip_bits(__deku_input.1)?;
+            let bytes = __deku_input.1 / 8;
+            let bits = __deku_input.1 % 8;
+            if bytes != 0 {
+                use ::#crate_::no_std_io::Seek;
+                use ::#crate_::no_std_io::SeekFrom;
+                if let Err(e) = __deku_reader.seek(SeekFrom::Current(i64::try_from(bytes).unwrap())) {
```
sharksforarms, line 139: same here

deku-derive/src/macros/deku_read.rs
```diff
@@ -121,9 +131,19 @@ fn emit_struct(input: &DekuData) -> Result<TokenStream, syn::Error> {
             use ::#crate_::DekuReader as _;
             let mut __deku_cursor = #crate_::no_std_io::Cursor::new(__deku_input.0);
             let mut __deku_reader = &mut deku::reader::Reader::new(&mut __deku_cursor);
-            if __deku_input.1 != 0 {
-                __deku_reader.skip_bits(__deku_input.1)?;
+            let bytes = __deku_input.1 / 8;
```
sharksforarms, line 134: I see this section is being repeated, worth having an `emit_..` function for this new chunk?

wcampbell0x2a: I was thinking that also, it's weird to codegen this code over and over again instead of something in the `reader`.

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: Yah, the `seek_*` code has the same problem.

sharksforarms (dismissed):
ty!

deku-derive/src/macros/deku_read.rs
wcampbell0x2a: Updated, I forgot skip_bits was already only used for this.

sharksforarms (approved):
nice!

PR #521, opened 2025-01-24
`Reader` shouldn't require its inner type to be a mutable reference

wgreenberg:
This makes `Reader` more general, so it can e.g. take ownership of a `Read + Seek` type.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -61,7 +61,7 @@ impl<R: Read + Seek> AsMut<R> for Reader<R> {
     }
 }
 
-impl<'a, R: Read + Seek> Reader<R> {
+impl<R: Read + Seek> Reader<R> {
     /// Create a new `Reader`
     #[inline]
     pub fn new(inner: R) -> Self {
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -32,8 +32,8 @@ enum Leftover {
 }
 
 /// Reader to use with `from_reader_with_ctx`
-pub struct Reader<'a, R: Read + Seek> {
-    inner: &'a mut R,
+pub struct Reader<R: Read + Seek> {
+    inner: R,
     /// bits stored from previous reads that didn't read to the end of a byte size
     leftover: Option<Leftover>,
     /// Amount of bits read after last read, reseted before reading enum ids
@@ -42,7 +42,7 @@ pub struct Reader<'a, R: Read + Seek> {
     pub bits_read: usize,
 }
 
-impl<R: Read + Seek> Seek for Reader<'_, R> {
+impl<R: Read + Seek> Seek for Reader<R> {
     #[inline]
     fn seek(&mut self, pos: SeekFrom) -> no_std_io::io::Result<u64> {
         #[cfg(feature = "logging")]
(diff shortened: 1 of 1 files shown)


wcampbell0x2a:
Seem plausible, I needed to do the following in [backhand](https://github.com/wcampbell0x2a/backhand) for this patch:
```diff
diff --git a/backhand/src/metadata.rs b/backhand/src/metadata.rs
index 89b4994..bf76adf 100644
--- a/backhand/src/metadata.rs
+++ b/backhand/src/metadata.rs
@@ -117,12 +117,13 @@ pub fn read_block<R: Read + Seek>(
     superblock: &SuperBlock,
     kind: &Kind,
 ) -> Result<Vec<u8>, BackhandError> {
-    let mut deku_reader = Reader::new(reader);
+    let mut deku_reader = Reader::new(&mut *reader);
     let metadata_len = u16::from_reader_with_ctx(&mut deku_reader, kind.inner.data_endian)?;

     let byte_len = len(metadata_len);
     tracing::trace!("len: 0x{:02x?}", byte_len);
     let mut buf = vec![0u8; byte_len as usize];
+    // deku_reader.as_mut().read_exact(&mut buf)?; // or this instead of the above
     reader.read_exact(&mut buf)?;

     let bytes = if is_compressed(metadata_len) {
diff --git a/backhand/src/squashfs.rs b/backhand/src/squashfs.rs
index b4420c5..2240d3c 100644
--- a/backhand/src/squashfs.rs
+++ b/backhand/src/squashfs.rs
@@ -225,7 +225,7 @@ impl<'b> Squashfs<'b> {
         kind: &Kind,
     ) -> Result<(SuperBlock, Option<CompressionOptions>), BackhandError> {
         // Parse SuperBlock
-        let mut container = Reader::new(reader);
+        let mut container = Reader::new(&mut *reader);
         let superblock = SuperBlock::from_reader_with_ctx(
             &mut container,
             (
```

wgreenberg:
ack, it turns out this is something of a compiler limitation: rust won't automatically reborrow generics, hence why you have to manually reborrow. one way to rewrite backhand's usage would be:

```rust
            let mut my_reader = Reader::new(reader);
            _ = u16::from_reader_with_ctx(&mut my_reader, ()).unwrap();

            let reader = my_reader.into_inner();
            reader.read_exact(&mut vec![0; 10]).unwrap();
```

but i understand if you think this is too obtuse

wcampbell0x2a:
> ack, it turns out this is something of a compiler limitation: rust won't automatically reborrow generics, hence why you have to manually reborrow. one way to rewrite backhand's usage would be:
> 
> ```rust
>             let mut my_reader = Reader::new(reader);
>             _ = u16::from_reader_with_ctx(&mut my_reader, ()).unwrap();
> 
>             let reader = my_reader.into_inner();
>             reader.read_exact(&mut vec![0; 10]).unwrap();
> ```
> 
> but i understand if you think this is too obtuse

I'm not against the change, it's technically less restrictive. Do you have a source for the "rust won't automatically reborrow generics"?, I've run into this before and never went any further into research of the problem.

sharksforarms:
I'm in favor of being more general. Satisfyingly, the tests also pass. Does this mean it's backwards compat? Any more tests that should be added? wdyt @wcampbell0x2a @wgreenberg ?

wcampbell0x2a:
> I'm in favor of being more general. Satisfyingly, the tests also pass. Does this mean it's backwards compat? Any more tests that should be added? wdyt @wcampbell0x2a @wgreenberg ?

I'm fine with this change. It's backwards compat for our tests, but it defiantly changes our public API, as above I had to change some of backhand.

sharksforarms (approved):

Issue #520, opened 2025-01-23
Support generic containers


vhdirk:
For the library I'm making, I also want to create bindings to other languages (like python). Mozilla's https://github.com/mozilla/uniffi-rs seems pretty promising. But...
It expects enum variants that aren't primitive types to be wrapped in an Arc. Would it make sense to have a blanket implementation of Reader and Writer for stuff like Arc, Rc and Cell?

wcampbell0x2a:
I think that's need fine, looks like a cool project!

PR #522, opened 2025-01-24
add arc impl

vhdirk:
This adds an impl for Arc, as mentioned in https://github.com/sharksforarms/deku/issues/520.

I've been thinking about making this more generic, but I don't have that much time. So I just copied the Box impl

### DIFF
--- a/src/impls/arc.rs
+++ b/src/impls/arc.rs
@@ -1,0 +1,151 @@
+use no_std_io::io::{Read, Seek, Write};
+use std::sync::Arc;
+
+use alloc::vec::Vec;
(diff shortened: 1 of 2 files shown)


wcampbell0x2a (dismissed):

vhdirk:
While this works, I think it'd probably only be useful if Mutex and RwLock would also be implemented

wcampbell0x2a (approved):

vhdirk:
Anything more that needs to be done here?

sharksforarms (approved):
Thanks!

PR #534, opened 2025-03-15
Restore last bit read during id_pat

wcampbell0x2a:
* Restore last leftover reading in reader for bit reads during id_pat


sharksforarms:
thanks! looks like some simple CI failures

sharksforarms (approved):

PR #538, opened 2025-03-24
Add missing imports of `alloc::format`

Serial-ATA:
Using the macros in `#[no_std]` wouldn't compile in some cases since `format!` wasn't being imported


wcampbell0x2a:
I would like to see tests so this doesn't happen again. But lgtm.

wcampbell0x2a (approved):

PR #540, opened 2025-03-29
Remove re-reading for ip_pat

### LINKED ISSUE #533: Enums which have variant `id_pat = "_"` ignore `BitSize` ctx
I have an enum like this:

```rust
#[deku(id_type = "u8", bits = 3)]
pub enum MSCm2RFClass {
    #[deku(id_pat = "0")] Class1,
    #[deku(id_pat = "1")] Class2,
    #[deku(id_pat = "2")] Class3,
    #[deku(id_pat = "3")] Class4,
    #[deku(id_pat = "4")] Class5,
    #[deku(id_pat = "_")] Other(u8),
}
```

Ideally, I'd want values 0-4 stored as their respective class values, and then the other remaining 3-bit values (5, 6, 7) to be stored as `Other(val)`. However, what actually happens is the `BitSize` context is ignored, an entire byte is read, and I end up with values like `Other(255)`.

The generated code looks something like this:

```rust
            let __deku_value = match &__deku_variant_id {
                0 => Self::Class1,
                1 => Self::Class2,
                2 => Self::Class3,
                3 => Self::Class4,
                4 => Self::Class5,
                _ => {
                    {
                        let lvl = ::log::Level::Trace;
                        if lvl <= ::log::STATIC_MAX_LEVEL && lvl <= ::log::max_level() {
                            ::log::__private_api::log(
                                format_args!("Reading: {0}.{1}", "MSCm2RFClass", "field_0"),
                                lvl,
                                &(
                                    "pycrate_rs::nas::generated::emm::emmtrackingareaupdaterequest",
                                    "pycrate_rs::nas::generated::emm::emmtrackingareaupdaterequest",
                                    ::log::__private_api::loc(),
                                ),
                                (),
                            );
                        }
                    };
                    let __deku___field_0 = {
                        let __deku_value = {
                            if let Err(e) = __deku_reader.seek_last_read() {
                                return Err(::deku::DekuError::Io(e.kind()));
                            }
                            <u8 as ::deku::DekuReader<
                                '_,
                                _,
                            >>::from_reader_with_ctx(__deku_reader, ())?
                        };
```

which passes `()` for the ctx in the `_` case.

wcampbell0x2a:
* Use the last read value for the id instead of jumping back and re-reading that value.
* Compare the endian, bits,and bytes amt when reading from an id. This is important when writing
  as without this check you end up with different bytes being written. More fields could
  be checked against this?
* Reader: Remove last_bits_read_amt and seek_last_read

Should close #533

### DIFF
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -530,6 +530,61 @@ struct FieldData {
 }
 
 impl FieldData {
+    pub fn any_field_set(&self) -> bool {
+        // NOTE: Ignore ident
+        let mut any_option_set = self.endian.is_some();
+
+        #[cfg(feature = "bits")]
+        {
+            any_option_set = any_option_set || self.bits.is_some();
+        }
+
+        any_option_set = any_option_set || self.bytes.is_some() || self.count.is_some();
+
+        #[cfg(feature = "bits")]
+        {
+            any_option_set = any_option_set || self.bits_read.is_some();
+        }
+
+        any_option_set = any_option_set
+            || self.bytes_read.is_some()
+            || self.until.is_some()
+            || self.map.is_some()
+            || self.ctx.is_some()
+            || self.update.is_some()
+            || self.reader.is_some()
+            || self.writer.is_some();
+
+        #[cfg(feature = "bits")]
+        {
+            any_option_set = any_option_set || self.pad_bits_before.is_some();
+        }
+
+        any_option_set = any_option_set || self.pad_bytes_before.is_some();
+
+        #[cfg(feature = "bits")]
+        {
+            any_option_set = any_option_set || self.pad_bits_after.is_some();
+        }
+
+        // NOTE: Ignore default
+        any_option_set = any_option_set
+            || self.pad_bytes_after.is_some()
+            || self.temp_value.is_some()
+            || self.cond.is_some()
+            || self.assert.is_some()
+            || self.assert_eq.is_some()
+            || self.seek_from_current.is_some()
+            || self.seek_from_end.is_some()
+            || self.seek_from_start.is_some()
+            || self.bit_order.is_some()
+            || self.magic.is_some();
+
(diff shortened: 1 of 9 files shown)


wcampbell0x2a:
> Compare the bits and bytes amt when reading from an id

Actually, I wonder that maybe I just use the id (bits, bytes, endian), and expect no attributes on the id_pat id storage.

wcampbell0x2a:
> > Compare the bits and bytes amt when reading from an id
> 
> Actually, I wonder that maybe I just use the id (bits, bytes, endian), and expect no attributes on the id_pat id storage.
> 
> (I don't have time currently to implement this)

Pushed an update to do this. We could actually force that *no* attributes are used on the special storage.

wcampbell0x2a:
Pushed updates to dis-allow any attributes for `id_pat` id storage.

This could be done in another mr, but the following should be checked and is not legal:
```rs
#[deku(id_type = "u8")]
enum DekuTest {
    #[deku(id_pat = "_")]
    VarDefault,
}
```

wcampbell0x2a:
> Pushed updates to dis-allow any attributes for `id_pat` id storage.
> 
> This could be done in another mr, but the following should be checked and is not legal:
> 
> ```rust
> #[derive(Debug, PartialEq, DekuRead, DekuWrite)]
> #[deku(id_type = "u8")]
> enum DekuTest {
>     #[deku(id_pat = "_")]
>     VarDefault,
> }
> ```

Nevermind, that wasn't that bad to fix: d4049d6

sharksforarms:
@wcampbell0x2a tag me when you'd like to get a review, also, maybe we should create a release after?

wcampbell0x2a:
> @wcampbell0x2a tag me when you'd like to get a review, also, maybe we should create a release after?

Sure! I found one bug with this release in my crates, I'll push and update for that and make some release notes.

sharksforarms (approved):
Thanks!

sharksforarms (approved):

PR #544, opened 2025-04-08
allow id_pat without storage

sharksforarms:
Allow the use of `id_pat` without a storage field. In the case where a storage field is specified, we use that. Else, we try the enum discriminant. Finally, we fail if we cannot determine what to use to write the id.

Address https://github.com/sharksforarms/deku/issues/533#issuecomment-2781201234


wcampbell0x2a (approved):
oh nice! lgtm

sharksforarms:
@wcampbell0x2a done

sharksforarms:
x-ref: https://github.com/sharksforarms/deku/pull/540

PR #548, opened 2025-04-13
Add Error and ErrorKind in no_std_io re-exports

RatCornu:
Hello! I'm making a library that needs both `deku` and `no_std_io`, and I need to convert `no_std_io2::io::Error` into `mycrate::Error`. Thus, as some structs and traits are already re-exported by `deku`, I think re-exporting also `Error` and `ErrorKind` is also a good idea to avoid asking `no_std_io_2` as an other dependency


RatCornu:
Thanks!

PR #555, opened 2025-05-16
fix: fix and test of conditional fields depending on conditional fields

goto40:
I observed that it was not possible to introduce a skippable temporary field (skip+cond) depending on another temporary skippable field.

The problem was, that for skippable fields, the temporary variable in added within an `if`, but again used later for another field depending on the first skippable field (see unittest).

Solution: Moving the variable outside the `if` solves the problem (in the diff you see a "moved" `if` expression, which makes the variable `fspec2_6` accessible later in the code):
image

Please have a look. I gratefully appreciate any feedback to incorporate in this PR.

- cargo test seems to work fine.
- I added a test to illustrate the problem and to prove that it is fixed...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+## Fixes
+- fix and test of conditional temporary fields depending on conditional temporary fields (did not compile before). See [unittest](./tests/test_attributes/test_temp_value_with_cond.rs).
+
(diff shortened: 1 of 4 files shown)


goto40:
Note: I also changed one unrelated file (`noseek.rs`) in order to fix some clippy issue. I can undo this, if you prefer to clean this up separately...

sharksforarms (approved):
Great! Thank you very much for a well described PR, fix, and test.

PR #565, opened 2025-05-27
Bugfix/signed values in bitfields

### LINKED ISSUE #564: Problems serializing signed values (negative values) in bitfields
When using signed types in bitfields I cannot serialize negative values (deserializing works).
With the example below I get `encode error: InvalidParam("bit size 16 of input is larger than bit requested size 10")`.

Any idea where to start? I suspect this can be solved maybe in library level (not in the macro?). Any hint is appreciated, then I could prepare a PR... **The changed code would check that for signed values the "unused" bits are all 0 or all 1 - unlike for unsigned values, where the "unsed" bits must all be 0.**
The error message ("input is larger than bit requested size") can clearly be allocated to a code location...

```rust
#[test]
fn check_signed_i10_decode_encode() {
    #[derive(Debug,PartialEq,Default,Clone,DekuRead,DekuWrite)]
    pub struct TestStruct {
        #[deku(bits="1")]
        pub a: bool,
        #[deku(pad_bits_before = "5", bits="10", endian = "big")]
        pub b: i16,
    }

    let buffer = vec![0b10000011,0b11111110];
    //                        ^^   ^^^^^^^^10 bits

    let ((remaining_bytes, offset), test_struct) = TestStruct::from_bytes((&buffer, 0)).expect("decoder error");
    // println!("{decoded_asterix:#?}");
    
    // everything consumed?
    assert_eq!(offset, 0);
    assert_eq!(remaining_bytes.len(), 0);

    // check content
    assert_eq!(test_struct.a, true);
    assert_eq!(test_struct.b, -2);

    // write back and check
    assert_eq!(buffer, test_struct.to_bytes().expect("encode error"))
}
```

goto40:
This PR fixes #564 

**Background**: Deku checks that the stored value fits into the specified bitfield.
**Problem**: There was a problem with checking the validity of signed integral values in bitfields: negative values have "1" set all along the unused area (for negative values; e.g. `-2u4 stored in a u8 = 0x1111 1110`). This caused a problem (deku checked for "1" in the unused region; see #564).

**Solution**:
* The logic was changed for signed types - separately for positive and negative values.
* I enhanced the macro `ImplDekuWrite` to accept an additional flag "Signed" or "Unsigned". All uses were adapted (importantly, for signd types we call the signed variant).
* I renamed some other macros accordingly (Signed/Unsigned) to make clear what is called/generated.
* I added a test for the different cases, as well, as for some trivial cases to see that it works as expected (min/max values and values one below minimum and one above maximum).

I am happy to get feedback (there is a coverage problem, see below).


goto40:
Hm. Any idea how I could trigger the remaining uncovered LOC? (see failed coverage analysis)

goto40:
I will refactor my changes and remove the clone code. This will probably solve the coverage issue.

goto40:
Still I encounter a problem: I do not know how to trigger the untouched error case:

image

goto40:
Do you have some advice how to trigger the missing parts?
I have nearly reached the goal in coverage: ` 88.76% of diff hit (target 88.85%)`

sharksforarms:
Hi @goto40 thanks for the PR! I wouldn't block merging on this requirement. It's more-so meant as a guideline. As long as tests are included which covers what we're changing, I'm not too worried! Taking a look at the PR.

wcampbell0x2a:
lgtm! thanks.

PR #570, opened 2025-07-08
Add map attribute to "list of attributes" in documentation

lieuwex:
The `map` attribute was missing from the "List of attributes". This commit adds it.


sharksforarms (approved):
:raised_hands:

PR #573, opened 2025-07-14
impls: hashset: tests: Handle arbitrary iteration order of HashSet

amboar:
FxHashSet is a type alias for a particular construction of std::collections::HashSet, whose iter() documentation states:

> An iterator visiting all elements in arbitrary order.

As such, we can't trust that the byte-sequence written to the writer buffer can be expected to match any particular sequence beyond the bytes internal to the write granule (u16).

The DekuWriter implementation for FxHashSet suggests switching to a deterministic hasher if a consistent output ordering is desired. However, this is still not sufficient for the circumstances, because while the output ordering might then be deterministic, there's no requirement that it match the input byte vector for the test at hand.

Chunk and search the expected_write vector for chunks of the write buffer.

The change is motivated by the observation of test failures on my Debian Bookworm system:

    > rustup show active-toolchain --verbose
    stable-x86_64-unknown-linux-gnu
    active because: it's the default toolchain
    compiler: rustc 1.88.0 (6b00bc388 2025-06-23)
    path: /home/andrew/.rustup/toolchains/stable-x86_64-unknown-linux-gnu

    > cargo test
    ...
    failures:

    ---- impls::hashset::tests::test_hashset_read_write::case_2_normal_be stdout ----

    thread 'impls::hashset::tests::test_hashset_read_write::case_2_normal_be' panicked at src/impls/hashset.rs:332:9:
    assertion `left == right` failed
      left: [170, 187, 204, 221]
     right: [204, 221, 170, 187]

    ---- impls::hashset::tests::test_hashset_write::case_1_normal stdout ----

    thread 'impls::hashset::tests::test_hashset_write::case_1_normal' panicked at src/impls/hashset.rs:286:9:
    assertion `left == right` failed
      left: [187, 170, 221, 204]
     right: [221, 204, 187, 170]

    failures:
        impls::hashset::tests::test_hashset_read_write::case_2_normal_be
        impls::hashset::tests::test_hashset_write::case_1_normal

    test result: FAILED. 175 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s


wcampbell0x2a:
We could also switch to an ordered mapping in our test code, but I think this is fine as a change also.

sharksforarms:
Let's go with this for now, thanks!

PR #575, opened 2025-07-16
Warnings and clippy fixes

amboar:
On the path of other efforts I came across a bunch of warnings/clippys. Here are some patches that tidy them up.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,7 +51,7 @@ jobs:
         with:
           toolchain: stable
 
-      - run: cargo clippy -- -D warnings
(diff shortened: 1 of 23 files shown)


sharksforarms:
Nice! Is there something we should change in the CI to catch these? additional flags, targets, etc?

amboar:
So my process wasn't super rigorous, lots of these came about via LSP in my editor. But perhaps we can add something in the vein of `RUSTFLAGS="-D warnings" cargo check --all-targets` to hit most of them (and more). Give me a moment.

wcampbell0x2a:
Just noting that this will create more CI churn whenever rust releases a new stable. Unless we pin the stable version.

PR #578, opened 2025-07-20
Implements From<DekuError> for no_std_io::io::Error in no_std context

RatCornu:
* Adds implementation of `From<DekuError>` for `no_std_io::io::Error` in a no_std context
* Removes `error_in_core` feature which does not seem necessary anymore


wcampbell0x2a (approved):
Seems fine, core::error::Error was added in 1.81 and that's our MSRV

PR #582, opened 2025-07-22
Expose `alloc` as a selectable feature

amboar:
Some environments prefer to avoid `alloc` in addition to building with `no_std`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,6 +68,7 @@ jobs:
           toolchain: nightly
           targets: thumbv7em-none-eabihf
       - run: cd ensure_no_std && cargo build --release --target thumbv7em-none-eabihf
+      - run: cd ensure_no_std && cargo build --release --target thumbv7em-none-eabihf --no-default-features
 
   ensure_wasm:
     name: Ensure wasm
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,10 +20,14 @@ jobs:
       - run: cargo build --all
       # test
       - run: cargo test --all
+      - run: cargo test --all-features
+      - run: cargo test --no-default-features
+      - run: cargo test --no-default-features --features=alloc
+      - run: cargo test --no-default-features --features=descriptive-errors
+      - run: cargo test --no-default-features --features=bits
+      - run: cargo test --no-default-features --features=logging
       # run examples
       - run: cargo run --example 2>&1 | grep -P '   ' | awk '{print $1}' | xargs -i cargo run --example {}
-      # test with no bits feature (don't test docs)
-      - run: cargo test --lib --examples --tests --features std --no-default-features
 
   # Only build on MSRV, since trybuild will fail on older version
   build-msrv:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Added
 
 - The `alloc` feature, allowing use in environments lacking a heap [#582](https://github.com/sharksforarms/deku/pull/582)
+- The `descriptive-errors` feature, replacing `no-assertion-string` [#582](https://github.com/sharksforarms/deku/pull/582)
 
 ## [0.19.1] - 2025-05-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+## Added
+
+- The `alloc` feature, allowing use in environments lacking a heap [#582](https://github.com/sharksforarms/deku/pull/582)
+
 ## [0.19.1] - 2025-05-22
 
 ## Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ members = [
 [features]
 default = ["std", "bits"]
 std = ["deku_derive/std", "bitvec?/std", "alloc", "no_std_io/std"]
-alloc = ["bitvec?/alloc"]
+alloc = ["bitvec?/alloc", "deku_derive/alloc", "no_std_io/alloc" ]
 logging = ["deku_derive/logging", "log"]
 no-assert-string = ["deku_derive/no-assert-string"]
 bits = ["dep:bitvec", "deku_derive/bits", "alloc" ]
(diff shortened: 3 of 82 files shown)


amboar:
Hello! This PR follows from the associated discussion thread:

https://github.com/sharksforarms/deku/discussions/580

I expect it will need (quite?) a few iterations - I'm happy to make progress by mining it for ideas that we can split  into separate patches or PRs if necessary.

wcampbell0x2a:
> https://github.com/sharksforarms/deku/pull/582/commits/552d6e7a0f1ccdee0db21d23a2ec54ce409d2fb1

It would be great if this commit has a passing test.

> https://github.com/sharksforarms/deku/pull/582/commits/cff993a1e51afe841eb2c6f64bab8eddca710580

I'm confused about the intent of this? I originally added this so that on embedded platforms you wouldn't end up with large string sections. And instead a simple return.

amboar:
> > [552d6e7](https://github.com/sharksforarms/deku/commit/552d6e7a0f1ccdee0db21d23a2ec54ce409d2fb1)
> 
> It would be great if this commit has a passing test.

I'll make a separate PR for it and add a test case.

> 
> > [cff993a](https://github.com/sharksforarms/deku/commit/cff993a1e51afe841eb2c6f64bab8eddca710580)
> 
> I'm confused about the intent of this? I originally added this so that on embedded platforms you wouldn't end up with large string sections. And instead a simple return.

That's a fair concern, because I don't think I described what I was actually up to there. The underlying motivation was around updating the CI workflow to allow `--all-features` to produce a "regular"-looking build, but the negative behaviour of the `no-assert-string` feature gets in the way.

I figured we could keep the intent of `no-assert-string` without requiring the feature by taking advantage of the `deku_error!()` macro from [21d8506](https://github.com/sharksforarms/deku/pull/582/commits/21d85062864f6c140579f5b335942030648e2629). However, I hadn't yet made use of the `assert` attribute to appreciate the impacts.

For something to experiment with I grabbed [backhand](https://github.com/wcampbell0x2a/backhand). As it's regular userspace I'm not sure it gives the greatest appreciation for the problem, but it's start.

Rather than pull out the `no-assert-string` feature entirely as I did in `cff993a` I [replaced it with `descriptive-errors`](https://gist.github.com/amboar/c0c36a8b0fe50f911bb417984dced9fc), which gives us a couple of things:

- What I feel is desirable behaviour if we pass `--all-features` to the build
- The ability to control the error message behaviour independent of the proposed `alloc` feature.

Here's the impact after switching backhand to my local deku tree:

```console
[I] 0 andrew@heihei ~/s/g/w/backhand (master)> git grep '^deku ='
backhand/Cargo.toml:deku = { path = "/home/andrew/src/github.com/sharksforarms/deku", default-features = false, features = ["std"] }
```

Given deku's features are explicitly specified by backhand, I force-enabled `descriptive-errors` in the build invocation to maintain current behaviour:
```console
[I] 0 andrew@heihei ~/s/g/w/backhand (master)> cargo build --release --features="deku/descriptive-errors"
    Finished `release` profile [optimized] target(s) in 0.06s
[I] 0 andrew@heihei ~/s/g/w/backhand (master)> bloaty -d sections target/release/unsquashfs-backhand | grep '[.]rodata$'
   8.5%   530Ki  10.7%   530Ki    .rodata
[I] 0 andrew@heihei ~/s/g/w/backhand (master)> readelf --string-dump .rodata target/release/unsquashfs-backhand | grep --color=never --only-matching -E "Field failed assertion.{5}"
Field failed assertion: Dir
Field failed assertion: Sup
Field failed assertion: Sup
Field failed assertion: Sup
Field failed assertion: Ext
Field failed assertion: Bas
Field failed assertion: Dir
```
For comparison, without enabling `descriptive-features`, I see:
```console
[I] 0 andrew@heihei ~/s/g/w/backhand (master)> cargo build --release
    Finished `release` profile [optimized] target(s) in 0.05s
[I] 0 andrew@heihei ~/s/g/w/backhand (master)> bloaty -d sections target/release/unsquashfs-backhand | grep '[.]rodata$'
   8.5%   529Ki  10.7%   529Ki    .rodata
[I] 0 andrew@heihei ~/s/g/w/backhand (master)> readelf --string-dump .rodata target/release/unsquashfs-backhand | grep --color=never --only-matching -E "Field failed assertion.{5}"
Field failed assertionevent
Field failed assertionCould
```
It appears the string coalescing behaves (mostly) as I'd expect, given we discard the non-constant portions of the strings in `deku_error!()` when disabling `descriptive-strings`.

I can squash the `descriptive-errors` patch into [cff993a](https://github.com/sharksforarms/deku/pull/582/commits/cff993a1e51afe841eb2c6f64bab8eddca710580) if you're okay with that approach and the analysis here (I feel it's okay, as it still gives us the control we need)?

wcampbell0x2a (dismissed):
lgtm!

wcampbell0x2a:
> @wcampbell0x2a thanks! Backing up slightly:
> 
> > I can squash the descriptive-errors patch into [cff993a](https://github.com/sharksforarms/deku/pull/582/commits/cff993a1e51afe841eb2c6f64bab8eddca710580) if you're okay with that approach and the analysis here (I feel it's okay, as it still gives us the control we need)?
> 
> I hadn't actually gone ahead with that as I was waiting on your feedback. However, I'll do so shortly given my impression that the discussion above was reasonable. Let me know otherwise.

Oh yes. I think I'm in favor of that patch.

amboar:
Okay, I've tidied up the `descriptive-errors` patch and squashed it into the earlier commit. A few new fixes  appear earlier in the series. I hit some new warnings, likely a result of upgrading to the latest rust stable release.

I've also updated the changelog in both [721e24e](https://github.com/sharksforarms/deku/pull/582/commits/721e24e235668ee5e93bf3bc25374a123351a726) and [2478ffc](https://github.com/sharksforarms/deku/pull/582/commits/2478ffcb3cc0686c21c0ff8c68098d6e3fbad3c7). Hopefully that's not a hindrance!

sharksforarms (approved):
Thanks! Looks like a good cleanup of some testing/errors too

sharksforarms (approved):

amboar:
Thanks @sharksforarms and @wcampbell0x2a !

sharksforarms:
Thank you for the contribution!

PR #588, opened 2025-07-28
Removed duplicate `bits` entry from the 'List of attributes' section of attributes docs

erius0:
The exact same entry is still present at line 44


sharksforarms (approved):
thanks!

PR #595, opened 2025-08-14
Handle bit order padding in Writer::finalize

maxpowa:
Noticed this while reverse engineering some networking code, my sample data seemed correct but the last byte would always get malformed when writing back. Turns out it was due to the alignment padding logic in `Writer::finalize` ignoring the previous writes' order.

- Corrected the padding logic in Writer::finalize to handle MSB0 and LSB0 bit orders separately.
- Updated and added tests to verify correct output for mixed bit orders and padding scenarios.


amboar:
> I don't know how to address the clippy errors, they don't occur when running clippy locally... `cargo clippy --all-targets -- -D warnings`

@maxpowa The first few changes in #582 address them. @sharksforarms @wcampbell0x2a I can split those out if you like.

maxpowa:
> @maxpowa The first few changes in #582 address them. @sharksforarms @wcampbell0x2a I can split those out if you like.

Looks like that was merged, and the clippy check is passing now!

sharksforarms:
Thanks for the PR and patience, I'll likely get to reviewing tomorrow or monday

sharksforarms (approved):
great find! thank you.

PR #597, opened 2025-09-06
Fix big endian bit_order lsb writing

wcampbell0x2a:
* Fix writing under 8 bits when Big Endian Lsb


sharksforarms:
thanks for your continued efforts @wcampbell0x2a. Time to cut a release you think?

Issue #607, opened 2025-10-11
Questioning requirement of alloc for read-only bitfield scenarios


nullstalgia:
Howdy, hope all is well.

`deku` has the most pleasant syntax for declaring structs as bitfields for encoding and decoding I've encountered so far!

Which makes me a little sad that an `alloc` requirements exists for all bitfields, regardless if you're reading or writing, and even if you have member types that are no larger than a `usize`.

I've spent a little time trying to dig through the master branch to find the uses of the core `alloc` and `bitvec` alloc types, and the vast majority seem to reside in tests and in Writer-related activities.

I do see *some* uses in the readers, but my naive assumption is that it seems like *some* BitVecs could be swapped with a BitSlice or similar?

Another curiosity was this comment in the Primitive reader: https://github.com/sharksforarms/deku/blob/257a1d4a94e60bdfb1cd8f7409b1554fb22f9d04/src/impls/primitive.rs#L194-L199

I'm curious about the source of the comment, if its as simple as "the BitField type doesn't support LSB0 reading with that cursed type of offset", fair enough, but the question of do we *still* need an allocator for decoding a `i32` encoded as `i14` is still ringing in my head.

I ask all this because I have several bitfield payloads with static/bounded sizes (~60 bytes each, give or take), all using plain integer or bool types with none individually larger than 32-bits. Despite having a deceptively simple simple use case, I'm unable to ship the deku-created parser without a full-fledged global allocator. 

I'd be happy to help bring a `no_alloc` read-only `deku` to fruition if the specific pain points are shown to me (if I don't dive in myself first).

amboar:
> Fixing the problems will likely need some input from ...

Scratch that, I've fixed it (slightly clunky) and dropped the patch disabling the tests. I can create a PR for the changes to pick over the approaches and style if people are happy for it.

PR #614, opened 2025-10-16
Lift alloc requirement for bits feature

amboar:
This has been of interest to me for a while. In parallel others had similar thoughts, e.g. in https://github.com/sharksforarms/deku/issues/607. I've opened the PR on the back of discussion there.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,7 +37,7 @@ jobs:
       matrix:
         toolchain:
             # msrv
-          - 1.81
+          - 1.82
 
     steps:
       - uses: actions/checkout@v4
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ keywords = ["deku", "bits", "serialization", "deserialization", "struct"]
 categories = ["encoding", "parsing", "no-std"]
 description = "bit level serialization/deserialization proc-macro for structs"
 readme = "README.md"
-rust-version = "1.81"
+rust-version = "1.82"
 
 [lib]
 bench = false
--- a/deku-derive/Cargo.toml
+++ b/deku-derive/Cargo.toml
@@ -7,7 +7,7 @@ license = "MIT OR Apache-2.0"
 repository = "https://github.com/sharksforarms/deku"
 description = "bit level serialization/deserialization proc-macro for structs"
 readme = "../README.md"
-rust-version = "1.81"
+rust-version = "1.82"
 
 [lib]
 proc-macro = true


amboar:
I hit a bug in the leftover bit-order swap path for the writer. I've fixed it in a fairly terrible way (exploded the code-paths so they weren't dependent, then fixed them independently). I think it should probably be tidied up for maintenance purposes, but I've pushed what I've got so the underlying bug is fixed. I've marked the PR as `[WIP]` as a result.

I don't believe I've done anything to trigger the MSRV failure, which seems to relate to indexmap. It looks like it had a release a couple of days ago, but yet to look into the problem in detail.

src/attributes.rs
```diff
@@ -220,7 +220,7 @@ With endian-ness:
 ```rust
 # use core::convert::{TryInto, TryFrom};
 # use deku::prelude::*;
-# #[cfg(feature = "bits")]
+# #[cfg(all(feature = "alloc", feature = "bits"))]
```
wcampbell0x2a: This shouldn't need alloc.

src/attributes.rs
```diff
@@ -245,14 +245,14 @@ let bytes = big_endian.to_bytes().unwrap();
...
-# #[cfg(not(feature = "bits"))]
+# #[cfg(not(all(feature = "alloc", feature = "bits")))]
 # fn main() {}
 ````
 ### Field Example
 ```rust
 # use core::convert::{TryInto, TryFrom};
 # use deku::prelude::*;
-# #[cfg(feature = "bits")]
+# #[cfg(all(feature = "alloc", feature = "bits"))]
```
wcampbell0x2a: This shouldn't need alloc.

src/attributes.rs
```diff
@@ -1334,7 +1334,7 @@ Example:
 # use alloc::{vec, vec::Vec};
 # use core::convert::{TryInto, TryFrom};
 # use deku::prelude::*;
-# #[cfg(feature = "bits")]
+# #[cfg(all(feature = "alloc", feature = "bits"))]
```
wcampbell0x2a: This shouldn't need alloc.

src/attributes.rs
```diff
@@ -1421,7 +1421,7 @@ Example:
...
```
wcampbell0x2a: This shouldn't need alloc.

wcampbell0x2a (changes requested):

amboar:
> Just to have the idea floating around, would it make sense for targets with allocators to use a heap-allocated BitVec for potentially "huge" types, and default to the stack-based BoundedBitVec for most types sized under whatever "huge" threshold?
> 
> https://github.com/sharksforarms/deku/pull/614/files#diff-ab13062c7a9940b962bf1fd3d9e8288d3effec084bc391b6c3945f893171df12R191
> 
> Or if there's no real benefit, that's fine.

Not that there's no real benefit, but maybe just some missing context: The specific case of the linked code implements the traits for Rust's primitive types. You can see the tree of macros invoking the implementations here:

https://github.com/amboar/deku/blob/bit-precise-no-alloc/src/impls/primitive.rs#L1209-L1260

As such there are no types that I'd consider "huge" there, in the sense of being inappropriate for the stack.

That said, the strategy of offering different implementations based on `cfg(feature = "alloc")` is reasonable but not something that's implemented yet.

src/attributes.rs
amboar: My intent was to keep it consistent with the config guard on `main()`, but you're right, and I'll drop it.

amboar:
@wcampbell0x2a I've addressed your comments so far, and have fixed the missing API docs as well.

tests/test_alloc.rs
```diff
@@ -66,7 +66,7 @@ mod tests {
                 let _ = TestDeku::from_reader((&mut cursor, 0)).unwrap();
             })
             .0,
-            (6, 0, 6)
+            (1, 0, 1)
```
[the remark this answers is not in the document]
amboar: This was satisfying! :D

Issue #599, opened 2025-09-08
Get payload size in a simpler way


redeexpressos:
hi. im trying to get the payload size using deku so I can write to a connection (asyncwrite). I dont have the `to_bytes` function in my message type so then I could use `len()`. (see `FIXME`)
Am I doing something wrong? I created the `create_packet` function to workaround but seems weird...

```rs
#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
pub struct KeepAlive {}

#[derive(Debug, Clone, Copy, PartialEq, DekuRead, DekuWrite)]
#[deku(endian = "endianness", ctx = "endianness: deku::ctx::Endian")]
pub struct MessageHeader {
    pub message_id: u32,
    pub len: u32,
}

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
#[deku(endian = "big")]
pub struct MessagePacket {
    pub header: MessageHeader,

    #[deku(ctx = "*header")]
    pub payload: Payload,
}

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
#[deku(ctx = "_: deku::ctx::Endian, header: MessageHeader")]
pub struct Payload {
    #[deku(
        bytes_read = "header.len",
        assert = "header.len <= 8192",
        ctx = "header.message_id"
    )]
    pub message_with_limit: MessageWithLimit,
}

#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
#[deku(id = "message_id", ctx = "message_id: u32")]
pub enum Message {
    #[deku(id = 0x0000_00FF)]
    KeepAlive(messages::KeepAlive),

   // ...
}


fn create_keep_alive_message() -> MessagePacket {
    let message = Message::KeepAlive(KeepAlive {});
    create_packet(message)
}

#[must_use]
fn create_packet(message: Message) -> MessagePacket {
    let message_id = message.deku_id().unwrap();

    // FIXME: weird solution to get payload size of Message to set remaining_bytes...??
    let mut buf = vec![];
    let mut cursor = Cursor::new(&mut buf);
    let mut writer = Writer::new(&mut cursor);
    message.to_writer(&mut writer, message_id).unwrap();
    let remaining_bytes = cursor.get_ref().len() as u32;

    let header = MessageHeader {  message_id, remaining_bytes };
    let message_with_limit = MessageWithLimit { message };
    let payload = ControlPayload { message_with_limit };

    MessagePacket { header, payload }
}


```

[...]

Issue #307, opened 2023-01-07
Add trait to get binary size of messages


wucke13:
If a message contains only known size types (e.g. arrays and primitives, but no `Vec` etc.) it would be cool to have a const helper function that determines the size of a buffer that is guaranteed to fit the serialized message. This greatly improves the usability in environments without alloc.

sempervictus:
Finding this to be somewhat necessary rather than a convenience when using typed enums getting their type from outside their own byte range (typed body getting type passed as ctx from the header) - can't convert those to bytes for measurement conveniently and need to edit them destructured which then has to update the header length.

wcampbell0x2a:
Should be closed with the addition of `DekuSize`.

PR #618, opened 2025-11-04
Add DekuSize impl

wcampbell0x2a:
* Add new derive DekuSize, which emits SIZE_BYTES and SIZE_BITS. This can be use to get the max bit/byte size of the parsed enum/struct so that your stack/heap can be properly allocated before usage.
* Generated for all primitive and std types we have, others such as Vec are impossible.

Closees #599, #307

### DIFF
--- a/src/attributes.rs
+++ b/src/attributes.rs
@@ -115,7 +115,13 @@ assert_eq!(data, &*value);
 # fn main() {}
 ```
 
-**Note**: The `endian` is passed as a context argument to sub-types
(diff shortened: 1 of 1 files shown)


sharksforarms:
Thanks! 

This should be ok given they're all types with known sizes. There may be some optimizations we could do for say `count = 5` on a `Vec<T>`

Can we have a trybuild test in order to confirm what the error is when we `DekuSize` on something like a Vec?

PR #657, opened 2026-05-06
fix: `Order::Lsb0` + `Endian::Little` bit alignment inconsistencies

### LINKED ISSUE #658: `Order::Lsb0` + `Endian::Little` alignment bug
## Issue

It appears that when a struct deriving `DekuRead` has a field that is one bit long and a field after that is large enough to cross byte boundaries that the `Endian::Little` context is not honored.

## Example

The testcase I created uses the bit pattern
```rust
// ...
     let bits = deku::bitvec::bits![
        u8, deku::bitvec::Lsb0;
        1, // Optional

        // Value - 24 bits. Should be evaluated as 255 when using Order::Lsb0 + Endian::Little
        1,1,1,1, 1,1,1,1,
        0,0,0,0, 0,0,0,0,
        0,0,0,0, 0,0,0,0,

        // Padding to make it byte aligned
        0,0,0,0, 0,0,0
    ]
// ...
```

and struct shape of

```rust
// ...
    #[deku(endian = "little", bit_order = "lsb")]
    struct TestStruct {
        #[deku(bits = 1)]
        optional: bool,
        #[deku(bits = 24)]
        value: u32,
    }
// ...
```

### Output

The output when parsing `TestStruct` is:
```
TestStruct {
    optional: true, // As expected
    value: 8323073  // <- Wrong
```

### Investigation

Converting the value we got (`8323073`) into binary we get the bit sequence `0b0111_1111_0000_0000_0000_0001`. It's less obvious with all but the first 8 bits being zero, but from my own troubleshooting it seems like this is where the parts of the final number are coming from.

```
[0][111_1111]_[0000_0000_0000_0001]
 ─ ─────────   ─────────────────── 
 ▲     ▲                      ▲    
 │┌────┘                      │    
 ││  1, // Field: `optional`  │    
 ││                           │    
 ││  // Field: `value`        │    
 │└─[1,1,1,1, 1,1,1],[1,      │    
 │   0,0,0,0, 0,0,0,0,        │    
 │   0,0,0,0, 0,0,0,0,]───────┘    
 │                                 
 │   // Padding                    
 └──[0],0,0,0 0,0,0                       
```

It's basically ignoring the remainder of byte 1 after reading the single bit for the field `optional`, using bytes 1 and 2 as expected (at least if they actually were the beginning of the value), appending the remaining bits from the byte 1 that was read and finally reading the last bit from byte 3 once it runs out of the remaining bits to pull from in byte 0.

---

Is this a known issue? It seems like there has been trouble in the past with LSB-LE parsing and misalignment but I couldn't find any open issues on the topic nor any mentions of known ongoing issues with it but I could have just missed it.

I created a draft pull request with a new testcase here -> #657

GrimOutlook:
Adding a test-case that I came across when trying to implement a message parsing standard.

Also added a possible fix for the issue.

fixes #658

### DIFF
--- a/src/impls/primitive.rs
+++ b/src/impls/primitive.rs
@@ -167,7 +167,7 @@ macro_rules! ImplDekuReadBits {
                 debug_assert!(MAX_TYPE_BITS >= bit_slice.len() + pad);
 
                 // if everything is aligned, just read the value
-                if pad == 0 && bit_slice.len() == MAX_TYPE_BITS {
+                if pad == 0 && bit_slice.len() == MAX_TYPE_BITS && order != Order::Lsb0 {
                     let bytes = bit_slice.domain().region().unwrap().1;
 
                     if bytes.len() * 8 == MAX_TYPE_BITS {
@@ -189,7 +189,7 @@ macro_rules! ImplDekuReadBits {
                 // Turning this into [0x23, 0x01] (then appending till type size)
                 debug_assert_eq!(bit_size, bit_slice.len());
                 const MAX_TYPE_BYTES: usize = core::mem::size_of::<$typ>();
-                if order == Order::Lsb0 && bit_slice.len() > 8 && pad != 0 {
+                if order == Order::Lsb0 {
                     let mut bits = crate::BoundedBitVec::<[u8; MAX_TYPE_BYTES], Msb0>::new();
 
                     bits.extend_from_bitslice(&bit_slice);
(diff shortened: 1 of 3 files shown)


GrimOutlook:
Since I'm unfamiliar with the codebase I had Gemini take a crack at fixing this by using the test case I made as a guide and it seems to have done a good job from what I can tell. I looked for any issues it might have created and I haven't been able to find anything glaring beyond some code duplication that could be cleaned up but moving the duplicated logic into the `read_bits_into` function breaks other calls to the function and tests start failing so I left it in. 

All tests pass outside of than some of the `test_compile` group but those seem to be failing in the previous CI runs as well so I don't believe this is caused by any changes in this branch.

wcampbell0x2a (approved):

PR #659, opened 2026-05-06
Avoid allocating in ReadExact when insufficient data

fbergero:
## Summary

- Adds a seek-based bounds check in `DekuReader<ReadExact> for Vec<u8>` to verify the underlying reader has enough bytes **before** allocating the output buffer
- Uses `reader.as_mut()` to seek on the inner reader directly, avoiding side effects on `Reader`'s `bits_read` and `leftover` state
- Returns `DekuError::Incomplete` early when data is insufficient

## Problem

The current implementation unconditionally allocates and zero-fills a buffer of `exact.0` bytes:

```rust
let mut bytes = alloc::vec![0x00; exact.0];
let _ = reader.read_bytes(exact.0, &mut bytes, Order::Lsb0)?;
```

If `exact.0` comes from untrusted input (e.g., a length-prefixed protocol), a malformed value can cause a huge allocation that gets immediately discarded when `read_bytes` fails. This is a DoS vector for any binary parser handling network data.

## Fix

Three cheap seeks (get position, seek to end, restore) to compute remaining bytes, then early-return `Incomplete` if insufficient. Zero overhead for the happy path beyond those seeks.

## Tests

- Unit tests for happy path, partial reads, empty input, zero-length, and error cases
- **Allocation-counting test** using `alloc_counter` (existing dev-dependency) that proves `(0, 0, 0)` allocations on the error path with `ReadExact(1_000_000)` on a 3-byte buffer

## Test plan

- [x] All existing vec tests pass (32/32)
- [x] New unit tests pass (6/6)
- [x] Allocation counting test confirms zero allocations on error path

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/impls/vec.rs
+++ b/src/impls/vec.rs
@@ -1,9 +1,10 @@
 use core::mem;
 
-use no_std_io::io::{Read, Seek, Write};
+use no_std_io::io::{Read, Seek, SeekFrom, Write};
 
 use alloc::vec::Vec;
 
+use crate::error::NeedSize;
 use crate::reader::Reader;
 use crate::writer::Writer;
 use crate::{ctx::*, DekuReader};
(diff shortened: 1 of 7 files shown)


fbergero:
@sharksforarms Hey idk what's your AI policy for contributions, but feel free to drop this PR if it's against the policy

wcampbell0x2a (approved):

Issue #602, opened 2025-09-15
Inconsistent parsing of structs with enums when using bit_order


zebrapurring:
When using `bit_order` on a struct that contains an enum, I noticed the result of parsing a byte array with `from_bytes` is different with respect to using a regular integer.

The following test reproduces the issue (tested on 0902ae06bd9612a70a4859dc75f007a5754ec54a):

```rs
use assert_hex::assert_eq_hex;
use deku::{ctx::Order, DekuContainerRead, DekuContainerWrite, DekuRead, DekuWrite};

#[derive(DekuRead, DekuWrite)]
#[deku(id_type = "u8", bits = 3)]
#[deku(ctx = "order: Order")]
pub enum MyEnum {
    #[deku(id = "0")]
    Val0,
    #[deku(id = "1")]
    Val1,
    #[deku(id = "2")]
    Val2,
    #[deku(id = "3")]
    Val3,
    #[deku(id = "4")]
    Val4,
    #[deku(id = "5")]
    Val5,
}

#[derive(DekuRead, DekuWrite)]
#[deku(bit_order = "lsb")]
pub struct StructWithEnum {
    #[deku(bits = 16)]
    f0: u16,
    f1: MyEnum,
    #[deku(bits = 13)]
    f2: u16,
}

#[derive(DekuRead, DekuWrite)]
#[deku(bit_order = "lsb")]
pub struct StructWithoutEnum {
    #[deku(bits = 16)]
    f0: u16,
    #[deku(bits = 3)]
    f1: u8,
    #[deku(bits = 13)]
    f2: u16,
}

fn main() {
    const INPUT: &[u8] = &[0xDE, 0xAD, 0xBE, 0xEF];
    let v1 = StructWithoutEnum::from_bytes((INPUT, 0)).unwrap().1;
    assert_eq_hex!(v1.to_bytes().unwrap(), INPUT, "❌ Non-enum variant failed");
    println!("✅ Non-enum variant passed");

    let v2 = StructWithEnum::from_bytes((INPUT, 0)).unwrap().1;
    assert_eq_hex!(v2.to_bytes().unwrap(), INPUT, "❌ Enum variant failed");
    println!("✅ Enum variant passed");
}
```

I would expect both variants to parse the same value, however the results are different:

```
✅ Non-enum variant passed

thread 'main' panicked at examples/enum.rs:50:5:
assertion `left == right` failed: ❌ Enum variant failed
  left: [
    0xde,
    0xad,
    0xfe,
[...]

PR #660, opened 2026-05-12
Auto-forward Order-typed ctx parameter as bit_order

sachaw:
Hi, I hit the following issue in one of my projects.
When a nested struct/enum receives bit ordering via `ctx = "o: Order"`, it currently has to also set `bit_order = "o"` for that order to be applied — otherwise the `ctx` parameter is silently ignored and codegen falls back to msb.
#602 hit this and was closed pointing to that duplicate-attribute workaround; this PR forwards an `Order`-typed `ctx` parameter as the runtime `bit_order` automatically when no explicit `bit_order` is given.

  ## Repro (against 0.20.3)

  ```rust
  use deku::prelude::*;

  #[deku(bit_order = "lsb")]
  struct Parent { left: Child, right: Child }

  #[repr(u8)]
  #[deku(bits = 4, id_type = "u8",
         ctx = "_o: deku::ctx::Order",
         ctx_default = "deku::ctx::Order::Lsb0")]
  enum Child { Empty = 0x0, Full = 0xf }

  #[test]
  fn lsb_propagates_from_ctx() {
      let v = Parent::try_from([0xf0].as_ref()).unwrap();
      assert_eq!(v, Parent { left: Child::Full, right: Child::Empty });
      // actual on master: Parent { right: Full, left: Empty } (msb read)
  }

### DIFF
--- a/deku-derive/src/lib.rs
+++ b/deku-derive/src/lib.rs
@@ -346,6 +346,28 @@ fn repr(attrs: &[Attribute]) -> Option<ReprType> {
     None
 }
 
+/// If `ctx` contains a parameter typed as `Order` (or `deku::ctx::Order`, `ctx::Order`),
+/// return the parameter name so it can be used as a runtime `bit_order` variable.
+#[cfg(feature = "bits")]
+fn find_order_param_in_ctx(
+    ctx: &syn::punctuated::Punctuated<syn::FnArg, syn::token::Comma>,
+) -> Option<String> {
+    for arg in ctx {
+        if let syn::FnArg::Typed(pat_type) = arg {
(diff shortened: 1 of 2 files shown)


amboar:
Not really a comment of substance, but:
```rust
struct Parent { right: Child, left: Child }
```
it seems like an odd choice to put the variable called `right` on the left, and `left` on the right

sachaw:
lol, good catch.

sachaw:
hey, any change I get get this merged, I have some other PR's to file afterwards.

wcampbell0x2a (approved):

wcampbell0x2a (changes requested):
Please fmt

wcampbell0x2a (approved):

PR #664, opened 2026-07-20
Fix id write endianness and sub-width signed sign-extension

gusanthon:
## Summary

Two independent correctness fixes for byte-aligned reads/writes, each with
regression tests. They're unrelated bugs that happened to surface together
while implementing a binary codec - happy to split into two PRs if you'd
prefer to review/merge them separately.

---

### Patch A - enum `id` written with the wrong endianness

**File:** `deku-derive/src/macros/deku_write.rs`

On the `id_pat` write path, `field_endian` was derived from `id_endian`
only, ignoring both a hardcoded `endian = "big"` and a bound ctx endian.
As a result an enum that declares `endian = "big"` would still write its
`id` in little-endian, so the written bytes didn't match the equivalent
flat big-endian field (and didn't round-trip against the read side).

**Fix:** mirror the read-side precedence already used in `gen_id_args`
(`id_endian` -> enum endian) for the `id_pat` write path, so the write side
resolves endianness the same way the read side does.

**Tests** (`tests/test_enum.rs`):
- an `id_pat` enum with `endian = "big"` writes the same bytes as the
  equivalent flat big-endian field
- a signed 24-bit `id_pat` enum round-trips negative sentinels on both LE
  and BE

---

### Patch B - sub-width signed reads not sign-extended

**File:** `src/impls/primitive.rs`

The `(Endian, ByteSize, Order)` `DekuReader` impl for signed types
(`ImplDekuReadSignExtend` macro) zero-padded the high bytes of a sub-width
byte-aligned read without propagating the sign bit. E.g. a 3-byte `i32`
read of `-100` returned `+16776092` instead of `-100`.

**Fix:** apply the same shift-based sign-extension already used in the
`BitSize` path, so the `ByteSize` path sign-extends consistently.

**Tests** (`src/impls/primitive.rs`): cover the `ByteSize` sign-extend path
directly for `i32` and `i16` sub-width reads on both endians.

### DIFF
--- a/deku-derive/src/macros/deku_write.rs
+++ b/deku-derive/src/macros/deku_write.rs
@@ -616,7 +616,9 @@ fn emit_field_write(
         // TODO: This should error on some attributes that don't make sense aren't used?
         // Such as magic, seek*
         let crate_ = super::get_crate_name();
-        let field_endian = input.id_endian.as_ref();
+        // Mirror gen_id_args' read-side precedence (id_endian -> enum endian) so a
+        // hardcoded `endian="big"` or a bound `endian="endian"` reaches the write.
+        let field_endian = input.id_endian.as_ref().or(input.endian.as_ref());
         #[cfg(feature = "bits")]
         let field_bits = input.bits.as_ref();
         #[cfg(not(feature = "bits"))]
--- a/src/impls/primitive.rs
+++ b/src/impls/primitive.rs
@@ -665,9 +665,15 @@ macro_rules! ImplDekuReadSignExtend {
                         };
                         // Sign-extend: a sub-width signed read zero-pads above, so the
                         // sign bit of the N-byte value must be propagated to the full type.
+                        // A zero-byte read has no sign bit to propagate, and shifting by
+                        // the full type width would overflow, so it is left as-is.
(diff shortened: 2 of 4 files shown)


wcampbell0x2a:
The following would panic with this patch:
```rust
#[derive(Debug, DekuRead)]
struct Dynamic {
    n: u8,
    #[deku(bytes = "*n as usize")]
    a: i32,
}
#[test]
fn dynamic_zero() {
    // n = 0 -> ByteSize(0) at runtime
    let mut c = Cursor::new([0x00u8, 0x01, 0x02]);
    println!("{:?}", Dynamic::from_reader((&mut c, 0)));
}


}
```

wcampbell0x2a (changes requested):
Requesting changes, see previous comment.

gusanthon:
Thanks for the review, @wcampbell0x2a, and good catch. Fixed in a4debfa by skipping the sign-extension when the size is 0. Your snippet gives `Ok((8, Dynamic { n: 0, a: 0 }))` again, same as master. Added it as a test in
`tests/test_regression.rs`. Also rebased onto master, so no more merge commits. Let me know if there's anything else that needs addressing.

wcampbell0x2a (approved):

PR #666, opened 2026-08-03
perf: zero-extend big-endian bit fields with load_be

### LINKED ISSUE #665: Reading a `bits = N` field is quadratic in the container width: the `insert(0, false)` pad loop
Reading a field declared `#[deku(bits = N)]` is much slower than it needs to be,
and the cost is driven by the *Rust type* the field is stored in rather than by how
many bits the field actually has. A big-endian 6-octet header with 11 bit-fields
(the CCSDS TM Transfer Frame primary header) takes **2.7 µs** to read.

The surprise is *where* the time goes. Only about 22% is the bit reader. The other
78% is a loop in `impls/primitive.rs` that prepends zero bits one at a time, where
each prepend shifts an entire array. Replacing that loop with
`BitField::load_be`, which `bitvec` already provides and which deku already uses a
few lines below in its `Lsb0` branch, makes the same header **4.0x** faster and
returns byte-identical values.

There is a self-contained reproduction at the bottom (one file, deku as the only
dependency). Every number below comes from its output.

## Background: the two steps in reading a `bits = N` field

Worth separating, because the cost is not in the step you would suspect.

1. **Get the bits out of the stream.** `Reader::read_bits_into` (`src/reader.rs`)
   pulls `N` bits and tracks a partial-byte "leftover" so the next field can
   continue mid-byte.
2. **Turn those `N` bits into the field's Rust type.** `DekuRead::read`
   (`src/impls/primitive.rs`) takes the `N`-bit slice and produces the `u8`,
   `u16`, ... value.

Three terms used throughout:

- **field width**: the `N` in `bits = N`.
- **container**: the Rust type the field is declared as, so `MAX_TYPE_BITS` is 8
  for `u8`, 16 for `u16`, and so on.
- **`pad`**: how many bits are missing to reach a whole number of bytes, that is
  `pad = 8 * ceil(N / 8) - N`.

Step 2 has to widen `N` bits into a `MAX_TYPE_BITS`-wide integer, which means
putting zero bits on the left (big-endian). That widening is where the time goes.

## The bug, by worked example

Take one real field: `#[deku(bits = 10)] scid: u16`, reading the 10 bits
`1010110101`. The answer we want is the `u16` `0b00000010_10110101`, or 693.

deku builds it in a `BoundedBitVec` whose backing store is a fixed
`BitArray<[u8; 2]>`, so 16 bits, the full width of the container:

```text
after extend_from_bitslice: the 10 field bits, left-aligned in a 16-bit array
   1010110101......                       size = 10

then `pad` = 6 zeros are prepended, ONE AT A TIME, and each
insert(0, false) shifts the WHOLE 16-bit array right by one bit:

   0 1010110101.....
   00 1010110101....
   000 1010110101...
   0000 1010110101..
   00000 1010110101.
   000000 1010110101                      size = 16

finally u16::from_be_bytes([0b00000010, 0b10110101]) = 693
```

So producing that value costs 6 whole-array shifts, about 96 bit moves, for
something `load_be` computes with a couple of word operations.

The relevant code, `src/impls/primitive.rs`:

```rust
// Some padding to next byte
let index = if input_is_le { bits.len() - (8 - pad) } else { 0 };
for _ in 0..pad {
    bits.insert(index, false);
}

// Pad up-to size of type
for _ in 0..(MAX_TYPE_BITS - bits.len()) {
    if input_is_le { bits.push(false); } else { bits.insert(0, false); }
}
```

and `BoundedBitVec::insert`, `src/lib.rs`:

```rust
fn insert(&mut self, index: usize, value: bool) {
    assert!(self.size < self.bits.len());
    assert!(index < self.size);
    let (_left, right) = self.bits.split_at_mut(index);
    right.shift_right(1);          // <-- shifts the whole backing array
    right.set(0, value);
    self.size += 1;
}
```

Two things make each `insert` expensive:

1. It shifts `self.bits`, the **entire** `BitArray<[u8; size_of::<$typ>()]>`, not
   just the `size` bits currently in use. So the work is proportional to the
   container width no matter how few bits the field has.
2. `split_at_mut` hands back an *aliased* `BitSlice` (`T::Alias`). `bitvec` cannot
   use its word-at-a-time `shift_right` on that, so it degrades to bit-by-bit.

Counting the two loops together, the number of `insert` calls is exactly
`MAX_TYPE_BITS - N`, and each one is `O(MAX_TYPE_BITS)`:

```text
work per field  =  O( container_bits x (container_bits - field_width) )
```

That is the whole bug. For `bits = 10` in a `u16` it is 6 shifts of 16 bits. For
`bits = 1` in a `u64` it is 63 shifts of 64 bits, and it measures 2.0 µs.

## Evidence

**The cost tracks the container, not the field.** The same single `bits = 1`
field, read into progressively wider containers. `cursor` is step 1
(`read_bits_into`) timed on its own, `conv` is step 2 by subtraction, and
`inserts` is the predicted `container_bits - 1`:

| type   |   total | cursor |    conv | inserts | ns/insert | via `load_be` |
| ------ | ------: | -----: | ------: | ------: | --------: | ------------: |
| `u8`   |  300 ns |  62 ns |  238 ns |       7 |      34.0 |         70 ns |
| `u16`  |  559 ns |  62 ns |  497 ns |      15 |      33.1 |         71 ns |
| `u32`  | 1011 ns |  62 ns |  949 ns |      31 |      30.6 |         70 ns |
| `u64`  | 2042 ns |  62 ns | 1979 ns |      63 |      31.4 |         70 ns |
| `u128` | 6795 ns |  62 ns | 6733 ns |     127 |      53.0 |         71 ns |

Step 1 is flat at 62 ns, as it should be, since it is the same one bit every time.
Step 2 grows 28x from `u8` to `u128` for that same one bit. The per-`insert` cost
is constant, so a single parameter fits the whole table:

```text
step 2  ~  33 ns x (container_bits - field_width)
```

`load_be` is flat, because it does not loop.

**Why this looks like "bit parsing is just slow".** There is an early return at
the top of `read` for the case `pad == 0 && N == MAX_TYPE_BITS`, meaning the field
exactly fills its container. That is why a plain `u8` field is fast, and why the
problem only ever shows up on `bits = N`. But the real trigger is `pad != 0`, not
sub-byte-ness. Six `u8` fields declared `bits = 8` against six bare `u8`s, same
wire layout, same alignment, only the code path differs:

```text
6x u8            12.2 ns   (read_bytes_const, no conversion)
6x bits=8 u8    103.2 ns   -> 8x
```

**Split for the 11-field header:**

```text
step 1, bit cursor            603.8 ns   (22%)
step 2, conversion           2130.9 ns   (78%)
```

## Suggested fix

`bitvec`'s `BitField::load_be` reads a bit slice as a big-endian integer and
zero-extends it. That is exactly the semantics the pad loop is hand-rolling, but
with word operations instead of a per-bit loop. deku already calls it in the
`Lsb0` branch just below. Adding a guarded fast path to both `DekuRead::read`
impls:

```rust
if !input_is_le && order == Order::Msb0 && bit_size > 0 {
    let value = bit_slice.load_be::<$inner>();
    return Ok((bit_size, <$typ>::from_be_bytes(value.to_be_bytes())));
}
```

Two notes on that line. The round-trip through `$inner` and `from_be_bytes` is
needed because this macro is also instantiated for `f32` and `f64`, which are not
`funty::Integral` and so cannot be `load_be` targets directly; for the integer
types the compiler folds it away. And `bit_size > 0` keeps today's behaviour for
`bits = 0`, which already panics inside `insert`.

Applied to 0.20.3:

| Check                                                     | Result                           |
| --------------------------------------------------------- | -------------------------------- |
| `cargo test --all-features --no-fail-fast`, unpatched      | 525 passed, 1 failed             |
| `cargo test --all-features --no-fail-fast`, patched        | 525 passed, 1 failed (identical) |
| `load_be` vs current conversion, `u8` to `u64`, all widths | 2.4M cases, 0 mismatches         |

The single failure is `test_compile`, a `trybuild` case whose expected stderr does
not match rustc 1.96. It fails identically with and without the patch.
`tests/bit_order.rs` (21 tests) and `tests/test_lsb_le.rs` (9 tests) stay green.

With a criterion bench added for big-endian bit-packed structs:

| Bench                            |  before |   after | change    |
| -------------------------------- | ------: | ------: | --------- |
| TM-shaped header, 11 fields      | 2782 ns |  701 ns | **3.97x** |
| one `bits = 1` field in a `u64`  | 2105 ns |   84 ns | **25x**   |
| six byte-aligned `u8`s (control) | 16.0 ns | 16.7 ns | no change |

## Scope, and a question

I restricted the fast path to explicit `endian = "big"` with the default `Msb0`
order, for two reasons:

- `Endian::default()` is the target CPU's endianness, so a struct that declares no
  endian gets little-endian on x86 and is **not** covered. That includes
  `DekuBits` in `benches/deku.rs`, which is why `deku_read_bits` barely moves
  (766 ns to 751 ns) and why I added a separate big-endian bench above.
- The little-endian and `Lsb0` branches compute something different, and #658 is
  an open correctness bug in exactly the `Lsb0 + Endian::Little` branch. Mixing a
  performance change into that seemed like a bad idea.

The little-endian case looks fixable the same way: assemble `ceil(N/8)` bytes with
the final partial byte right-aligned, then combine little-endian, which is
O(bytes) rather than O(bits^2). But that needs a decision on the intended
semantics first, given #658.

So: would you prefer the narrow big-endian and `Msb0` fast path, or a version that
covers little-endian and `Lsb0` too? Happy to open a PR with the narrow one plus
the big-endian bench, and follow up on the rest.

Related: #25 (benchmarks and performance), #658 (`Lsb0` + `Endian::Little`).

## Reproducing

Two files, deku as the only dependency, no criterion.

```toml
# Cargo.toml
[package]
name = "deku-bits-repro"
version = "0.0.0"
edition = "2021"

[dependencies]
deku = "0.20.3"

[profile.release]
opt-level = 3
```

<details>
<summary><code>src/main.rs</code></summary>

```rust
//! Minimal reproduction: reading a `bits = N` field is quadratic in the
//! container width. Only dependency is deku itself. `cargo run --release`.

use std::hint::black_box;
use std::time::Instant;

use deku::bitvec::{BitArray, BitField, Msb0};
use deku::ctx::{BitSize, Endian, Order};
use deku::no_std_io::Cursor;
use deku::prelude::*;
use deku::reader::Reader;

const BUF: &[u8] = &[0x2A, 0xB5, 0x11, 0x22, 0xC7, 0xFF, 0x00, 0x99, 0x44, 0x33, 0x88, 0x77];

/// Best-of-5 mean ns per call.
fn ns<F: FnMut() -> u64>(mut f: F) -> f64 {
    let mut w = 0u64;
    for _ in 0..20_000 {
        w = w.wrapping_add(f());
    }
    black_box(w);
    let mut best = f64::MAX;
    for _ in 0..5 {
        let n = 100_000;
        let mut acc = 0u64;
        let t = Instant::now();
        for _ in 0..n {
            acc = acc.wrapping_add(f());
        }
        let e = t.elapsed().as_secs_f64();
        black_box(acc);
        best = best.min(e * 1e9 / n as f64);
    }
    best
}

/// CCSDS TM Transfer Frame primary header: 6 octets, 11 fields, big-endian.
#[derive(DekuRead)]
#[deku(endian = "big")]
struct Tm {
    #[deku(bits = 2)]
    tfvn: u8,
    #[deku(bits = 10)]
    scid: u16,
    #[deku(bits = 3)]
    vcid: u8,
    #[deku(bits = 1)]
    ocf: u8,
    mcfc: u8,
    vcfc: u8,
    #[deku(bits = 1)]
    tfs: u8,
    #[deku(bits = 1)]
    syn: u8,
    #[deku(bits = 1)]
    po: u8,
    #[deku(bits = 2)]
    sli: u8,
    #[deku(bits = 11)]
    fhp: u16,
}

/// Six bare `u8`s: byte-aligned fast path.
#[derive(DekuRead)]
#[deku(endian = "big")]
struct Bare6 {
    a: u8,
    b: u8,
    c: u8,
    d: u8,
    e: u8,
    f: u8,
}

/// IDENTICAL wire layout to `Bare6`, but each field declared `bits = 8`.
#[derive(DekuRead)]
#[deku(endian = "big")]
struct Bits8x6 {
    #[deku(bits = 8)]
    a: u8,
    #[deku(bits = 8)]
    b: u8,
    #[deku(bits = 8)]
    c: u8,
    #[deku(bits = 8)]
    d: u8,
    #[deku(bits = 8)]
    e: u8,
    #[deku(bits = 8)]
    f: u8,
}

/// The same 11 fields, same bit cursor, but `load_be` instead of the pad loop.
fn tm_via_load_be() -> f64 {
    ns(|| {
        let mut r = Reader::new(Cursor::new(black_box(BUF)));
        let mut acc = 0u64;
        for w in [2usize, 10, 3, 1, 8, 8, 1, 1, 1, 2, 11] {
            let mut a: BitArray<[u8; 16], Msb0> = BitArray::ZERO;
            r.read_bits_into(&mut a.as_mut_bitslice()[..w], Order::Msb0)
                .unwrap();
            acc ^= a.as_bitslice()[..w].load_be::<u64>();
        }
        acc
    })
}

fn parse<T: for<'a> DekuReader<'a, ()>>(f: impl Fn(T) -> u64) -> f64 {
    ns(|| {
        let mut r = Reader::new(Cursor::new(black_box(BUF)));
        f(T::from_reader_with_ctx(&mut r, ()).unwrap())
    })
}

/// Cost of the bit cursor alone: `read_bits_into` with no integer conversion.
fn cursor(w: usize) -> f64 {
    ns(|| {
        let mut r = Reader::new(Cursor::new(black_box(BUF)));
        let mut a: BitArray<[u8; 16], Msb0> = BitArray::ZERO;
        r.read_bits_into(&mut a.as_mut_bitslice()[..w], Order::Msb0)
            .unwrap();
        a.as_raw_slice()[0] as u64
    })
}

/// One `bits = 1` field read into container `$t`, and the same via `load_be`.
macro_rules! one_bit {
    ($t:ty, $bits:expr) => {{
        let total = ns(|| {
            let mut r = Reader::new(Cursor::new(black_box(BUF)));
            <$t>::from_reader_with_ctx(&mut r, (Endian::Big, BitSize(1))).unwrap() as u64
        });
        let fixed = ns(|| {
            let mut r = Reader::new(Cursor::new(black_box(BUF)));
            let mut a: BitArray<[u8; 16], Msb0> = BitArray::ZERO;
            r.read_bits_into(&mut a.as_mut_bitslice()[..1], Order::Msb0)
                .unwrap();
            a.as_bitslice()[..1].load_be::<$t>() as u64
        });
        (stringify!($t), $bits, total, fixed)
    }};
}

fn main() {
    println!("\n1. A 6-octet header with 11 bit-fields, big-endian");
    let t_tm = parse::<Tm>(|v| {
        v.tfvn as u64
            ^ v.scid as u64
            ^ v.vcid as u64
            ^ v.ocf as u64
            ^ v.mcfc as u64
            ^ v.vcfc as u64
            ^ v.tfs as u64
            ^ v.syn as u64
            ^ v.po as u64
            ^ v.sli as u64
            ^ v.fhp as u64
    });
    let t_fix = tm_via_load_be();
    println!("   deku today                {t_tm:>9.1} ns");
    println!(
        "   same cursor + load_be     {t_fix:>9.1} ns   -> {:.1}x",
        t_tm / t_fix
    );

    println!("\n2. Same wire layout, byte-aligned: only the code path differs");
    let t_bare = parse::<Bare6>(|v| v.a as u64 ^ v.f as u64);
    let t_b8 = parse::<Bits8x6>(|v| v.a as u64 ^ v.f as u64);
    println!("   6x u8            {t_bare:>9.1} ns   (read_bytes_const)");
    println!("   6x bits=8 u8     {t_b8:>9.1} ns   -> {:.0}x", t_b8 / t_bare);
    println!("   => the trigger is `pad != 0`, not sub-byte-ness.");

    println!("\n3. Cost tracks the CONTAINER, not the field: one bits=1 field");
    println!(
        "   {:<6}{:>9}{:>10}{:>10}{:>9}{:>11}{:>10}",
        "type", "total", "cursor", "conv", "inserts", "ns/insert", "load_be"
    );
    let c1 = cursor(1);
    for (name, bits, total, fixed) in [
        one_bit!(u8, 8),
        one_bit!(u16, 16),
        one_bit!(u32, 32),
        one_bit!(u64, 64),
        one_bit!(u128, 128),
    ] {
        let conv = total - c1;
        let inserts = bits - 1;
        println!(
            "   {name:<6}{total:>8.0}n{c1:>9.0}n{conv:>9.0}n{inserts:>9}{:>11.1}{fixed:>9.0}n",
            conv / inserts as f64
        );
    }
    println!("   => conv ~ 33 ns x (container_bits - field_bits); load_be is flat.");

    println!("\n4. Where the header's time goes");
    let cur_sum: f64 = [2usize, 10, 3, 1, 8, 8, 1, 1, 1, 2, 11].iter().map(|w| cursor(*w)).sum();
    println!("   bit cursor, 11 fields    {cur_sum:>9.1} ns   ({:.0}%)", 100.0 * cur_sum / t_tm);
    println!(
        "   conversion (by subtr.)   {:>9.1} ns   ({:.0}%)",
        t_tm - cur_sum,
        100.0 * (t_tm - cur_sum) / t_tm
    );

    println!("\n5. load_be is byte-identical to the current conversion");
    let mut s = 0x9E37_79B9_7F4A_7C15u64;
    let (mut n, mut bad) = (0u64, 0u64);
    for _ in 0..20_000 {
        s ^= s >> 12;
        s ^= s << 25;
        s ^= s >> 27;
        let v = s.wrapping_mul(0x2545_F491_4F6C_DD1D);
        let mut buf = [0u8; 16];
        buf[..8].copy_from_slice(&v.to_be_bytes());
        buf[8..].copy_from_slice(&v.rotate_left(17).to_be_bytes());
        macro_rules! chk {
            ($t:ty, $max:expr) => {
                for w in 1..=$max {
                    let mut r = Reader::new(Cursor::new(buf.as_slice()));
                    let a = <$t>::from_reader_with_ctx(&mut r, (Endian::Big, BitSize(w))).unwrap();
                    let mut r2 = Reader::new(Cursor::new(buf.as_slice()));
                    let mut arr: BitArray<[u8; 16], Msb0> = BitArray::ZERO;
                    r2.read_bits_into(&mut arr.as_mut_bitslice()[..w], Order::Msb0).unwrap();
                    let b = arr.as_bitslice()[..w].load_be::<$t>();
                    if a != b {
                        bad += 1;
                    }
                    n += 1;
                }
            };
        }
        chk!(u8, 8);
        chk!(u16, 16);
        chk!(u32, 32);
        chk!(u64, 64);
    }
    println!("   {n} (type, width, input) cases checked, {bad} mismatches");
    println!();
}
```

</details>

Then:

```
cargo run --release
```

Output on deku 0.20.3, rustc 1.96.0, `opt-level = 3`, no LTO, AMD EPYC 9374F,
single thread:

```
1. A 6-octet header with 11 bit-fields, big-endian
   deku today                   2734.7 ns
   same cursor + load_be         673.1 ns   -> 4.1x

2. Same wire layout, byte-aligned: only the code path differs
   6x u8                 12.2 ns   (read_bytes_const)
   6x bits=8 u8         103.2 ns   -> 8x
   => the trigger is `pad != 0`, not sub-byte-ness.

3. Cost tracks the CONTAINER, not the field: one bits=1 field
   type      total    cursor      conv  inserts  ns/insert   load_be
   u8         300n       62n      238n        7       34.0       70n
   u16        559n       62n      497n       15       33.1       71n
   u32       1011n       62n      949n       31       30.6       70n
   u64       2042n       62n     1979n       63       31.4       70n
   u128      6795n       62n     6733n      127       53.0       71n
   => conv ~ 33 ns x (container_bits - field_bits); load_be is flat.

4. Where the header's time goes
   bit cursor, 11 fields        603.8 ns   (22%)
   conversion (by subtr.)      2130.9 ns   (78%)

5. load_be is byte-identical to the current conversion
   2400000 (type, width, input) cases checked, 0 mismatches
```

Simon-Calbert-Aerospacelab:
Closes #665.

Reading a `#[deku(bits = N)]` field zero-extends it to the container width with:

```rust
for _ in 0..(MAX_TYPE_BITS - bits.len()) {
    bits.insert(0, false);
}
```

`BoundedBitVec::insert` shifts the whole backing array right by one bit per call,
bit by bit, because the `split_at_mut` alias defeats bitvec's word-at-a-time
`shift_right`. That makes a field `O(container_bits * (container_bits - N))` bit
operations: a 1-bit field in a `u64` pays 63 whole-array shifts, about 2 us.

`BitField::load_be` does the same zero-extension in one step, and this file
already uses it in the `Lsb0` branch below.

## Change

Two hunks in `src/impls/primitive.rs`, one per `DekuRead::read` impl
(`(Endian, BitSize, Order)` and `(Endian, BitSize)`):

```rust
if !input_is_le && order == Order::Msb0 && bit_size > 0 {
    let value = bit_slice.load_be::<$inner>();
    return Ok((bit_size, <$typ>::from_be_bytes(value.to_be_bytes())));
}
```

`f32` and `f64` are not `funty::Integral`, so the value goes through the unsigned
`$inner` and `from_be_bytes`, as the existing padded-array path does.
`bit_size > 0` leaves `bits = 0` panicking in `insert` as before.

Adds `benches/bebits.rs`, since the existing `DekuBits` bench declares no endian
and never reaches this path.

## Scope

Explicit `endian = "big"` with the default `Msb0` only. Untouched: no `endian`
attribute (which defaults to target endianness, so little-endian on x86),
`endian = "little"`, and any `bit_order = "lsb"`.

I left the little-endian side alone on purpose. #658 is an open correctness bug in
the `Lsb0 + Endian::Little` branch, so mixing a performance change into it seemed
unwise. It is fixable the same way, building `ceil(N/8)` bytes with the final
partial byte right-aligned and combining little-endian, once the intended
semantics are settled. Happy to follow up.

## Testing

The full CI feature matrix passes and all nine examples run. `tests/bit_order.rs`
and `tests/test_lsb_le.rs` stay green.

A differential sweep of the old conversion against the new one, over `u8`, `u16`,
`u32` and `u64`, every valid width and 40k random inputs: 4.8M cases, 0
mismatches.

Two jobs are already red on unpatched `master`, so they are not from this PR:

- `test_compile`: the trybuild expected stderr does not match newer rustc.
  Verified identical by reverting only `primitive.rs`.
- `cargo clippy -- -D warnings`: `src/lib.rs:983` calls `shift_right`, deprecated
  in bitvec 1.1.1, which the `"1.0.1"` requirement now resolves to. Can fix
  separately if you want.

## Performance

`cargo bench --all-features --bench bebits`, against a `master` baseline:

| Bench                                | master  | patched | change         |
| ------------------------------------ | ------: | ------: | -------------- |
| `be_tm_primary_header_11_fields`     | 2827 ns |  712 ns | **-74.7%**     |
| `be_one_bit_in_u64`                  | 2050 ns | 79.1 ns | **-96.1%**     |
| `be_six_bytes_aligned` (byte-aligned control) | 15.9 ns | 15.8 ns | none, p = 0.19 |

`--bench deku` is unchanged: `deku_read_bits` 740 to 729 ns (p = 0.34). It did
report +2.2% on `deku_read_byte`, but the same binary run twice moves that one
-3.0%, so run-to-run drift on an 8 ns benchmark is larger than the effect.

rustc 1.96.0, opt-level 3, no LTO, AMD EPYC 9374F.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,11 @@ name = "deku"
 harness = false
 required-features = ["alloc"]
 
+[[bench]]
+name = "bebits"
+harness = false
+required-features = ["alloc", "bits"]
+
 [lints]
 workspace = true
 
--- a/benches/bebits.rs
+++ b/benches/bebits.rs
@@ -1,0 +1,75 @@
+//! Big-endian bit-packed headers: the shape used by real network / space
+//! protocols (CCSDS, IPv4, DVB-S2). Mirrors the CCSDS TM Transfer Frame
+//! primary header, 6 octets / 11 fields.
+use criterion::{criterion_group, criterion_main, Criterion};
+use deku::prelude::*;
+use no_std_io::io::Cursor;
+
+#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+#[deku(endian = "big")]
+struct TmPrimaryHeader {
+    #[deku(bits = 2)]
+    tfvn: u8,
+    #[deku(bits = 10)]
+    scid: u16,
+    #[deku(bits = 3)]
+    vcid: u8,
+    #[deku(bits = 1)]
+    ocf: u8,
+    mcfc: u8,
+    vcfc: u8,
+    #[deku(bits = 1)]
+    tfs: u8,
+    #[deku(bits = 1)]
+    syn: u8,
+    #[deku(bits = 1)]
+    po: u8,
+    #[deku(bits = 2)]
+    sli: u8,
+    #[deku(bits = 11)]
+    fhp: u16,
+}
+
+/// Same 6 octets, byte-aligned: the deku fast path, for scale.
+#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+#[deku(endian = "big")]
+struct SixBytes {
+    a: u8,
+    b: u8,
+    c: u8,
+    d: u8,
+    e: u8,
+    f: u8,
+}
+
+/// Worst case in the docs: a 1-bit field in a wide container.
+#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+#[deku(endian = "big")]
+struct OneBitU64 {
+    #[deku(bits = 1)]
+    a: u64,
+}
+
+fn bench(c: &mut Criterion) {
+    let buf = [0x2Au8, 0xB5, 0x11, 0x22, 0xC7, 0xFF, 0x00, 0x99];
+    c.bench_function("be_tm_primary_header_11_fields", |b| {
+        b.iter(|| {
+            let mut r = Reader::new(Cursor::new(&buf));
+            TmPrimaryHeader::from_reader_with_ctx(&mut r, ()).unwrap()
+        })
+    });
+    c.bench_function("be_six_bytes_aligned", |b| {
+        b.iter(|| {
+            let mut r = Reader::new(Cursor::new(&buf));
+            SixBytes::from_reader_with_ctx(&mut r, ()).unwrap()
+        })
+    });
+    c.bench_function("be_one_bit_in_u64", |b| {
+        b.iter(|| {
+            let mut r = Reader::new(Cursor::new(&buf));
+            OneBitU64::from_reader_with_ctx(&mut r, ()).unwrap()
+        })
+    });
+}
+criterion_group!(bebits, bench);
+criterion_main!(bebits);
--- a/src/impls/primitive.rs
+++ b/src/impls/primitive.rs
@@ -181,6 +181,22 @@ macro_rules! ImplDekuReadBits {
                     }
                 }
 
+                // Fast path: big-endian, Msb0 bit order.
+                //
+                // The generic paths below zero-extend the value to the container
+                // width by calling `BoundedBitVec::insert(0, false)` once per
+                // padding bit, and each `insert` shifts the entire backing array
+                // right by one bit. That is O(MAX_TYPE_BITS) bit operations per
+                // padding bit, i.e. O(MAX_TYPE_BITS * (MAX_TYPE_BITS - bit_size))
+                // per field -- for a 1-bit field in a u64 that is 63 whole-array
+                // shifts. `load_be` performs the identical zero-extension in one
+                // step. Only Msb0 + big-endian is handled here; the Lsb0 and
+                // little-endian paths keep their existing behaviour.
+                if !input_is_le && order == Order::Msb0 && bit_size > 0 {
+                    let value = bit_slice.load_be::<$inner>();
+                    return Ok((bit_size, <$typ>::from_be_bytes(value.to_be_bytes())));
+                }
+
                 // if read from Lsb order and it's especially cursed since its not just within one byte...
                 // read_bits returned: [0, 0, 0, 1, 0, 0, 1, 0, 0, 0, 1, 1]
                 //                     | second     |    first            |
@@ -284,6 +300,15 @@ macro_rules! ImplDekuReadBits {
                         };
                         return Ok((bit_size, value));
                     }
+                }
+
+                // Fast path: big-endian. This impl has no `Order` parameter and
+                // so is always Msb0. See the comment on the same fast path in
+                // the `(Endian, BitSize, Order)` impl above for why the generic
+                // path below is quadratic in the container width.
+                if !input_is_le && bit_size > 0 {
+                    let value = bit_slice.load_be::<$inner>();
+                    return Ok((bit_size, <$typ>::from_be_bytes(value.to_be_bytes())));
                 }
 
                 // Create a new BoundedBitVec from the slice and pad un-aligned chunks


wcampbell0x2a:
I bet we can do something like this for LE+LSB.

Simon-Calbert-Aerospacelab:
Good instinct, and I tried it before replying. It does not carry over, and the
reason turned out to be worth reporting.

This PR fixes a **padding** problem. With big-endian + Msb0 the bits arrive in
the order the number wants them, so the only work left is putting zeros in
front, and `load_be` is exactly that operation. That is why the swap is
provably equivalent rather than just close.

Lsb0 is not a padding problem. It needs a byte **reordering** (the existing code
walks bytes backwards with `rchunks_exact(8)` while reading the bits inside each
one forwards), and neither `load_be` nor `load_le` can express that.

I also benchmarked where the pad loop actually bites, by reading a 1-bit field
into containers of increasing width:

| container | BE+Msb0 (this PR) | BE+Lsb0 | LE+Msb0 | LE+Lsb0 |
|---|---|---|---|---|
| `u8`  | 77.6 ns | 323 ns  | 300 ns | 316 ns |
| `u64` | 79.7 ns | 2123 ns | 373 ns | 373 ns |

Little-endian is flat, which matches the source: the pad loop appends with
`push` when the input is little-endian and only prepends with `insert(0, ..)`
when it is big-endian. So **LE never had the #665 behaviour**, which is probably
why the dual felt forced. LE is still ~4.7x slower than this PR's path, but that
is a flat overhead in the generic assembly path, a different problem.

The quadrant that does still have the #665 defect is **BE+Lsb0**: 27x above, and
70x for a byte-aligned 16-bit field in a `u64` (22.5 ns vs 1572 ns). Dropping
`order == Order::Msb0` from the guard breaks `test_bit_order_more_first_be` and
`test_idempotency_multi_byte`, for the reordering reason above.

I would hold off on that quadrant anyway: #657's `skip` correction is
big-endian specific and it widens the branch to all Lsb0, so it changes BE+Lsb0
output too. Optimizing there today means validating against behaviour that is
about to change.

Happy to take BE+Lsb0 as a follow-up once #657 lands, with the same
bit-identical sweep I used here. Suggest keeping this PR scoped to BE+Msb0, the
one quadrant where the equivalence is provable today.

wcampbell0x2a (approved):

Simon-Calbert-Aerospacelab:
Follow-up worth flagging: I applied #657 on top of this branch, and it removes
the #665 defect from the Lsb0 column by itself.

The reason is that the expensive loop lives in the *generic* branch, the one
that pads all the way up to the container width with `insert(0, false)`. Today
Lsb0 only reaches the cheaper Lsb0-specific branch when
`len > 8 && pad != 0`, so everything else falls through to it. #657 widens that
condition to plain `order == Order::Lsb0`, and the fall-through stops.

Measured on top of this PR:

| BE+Lsb0 case | before #657 | with #657 |
|---|---|---|
| 1-bit field in a `u64` | 2123 ns | 374 ns |
| byte-aligned 16-bit field in a `u64` | 1572 ns | 62.9 ns |

The two patches apply cleanly together and pass `bit_order` 21/21 and
`test_lsb_le` 9/9.

So BE+Lsb0 might need no performance patch of its own, and these two can merge
independently. Once both are in, the #665 defect (quadratic complexity over the container size) is gone from every quadrant,
and what remains is a uniform ~4.7x from the `BoundedBitVec` assembly itself.

Simon-Calbert-Aerospacelab:
@wcampbell0x2a FYI I have two follow-up PRs sitting on my fork that take the
same BE + `Msb0` path further. Happy to open them here once this one lands.

1. **Read and write bit fields as integers.** This PR made the zero-extension
   cheap. That one goes a level up and skips the `BitSlice` entirely: the field
   is read straight into a `u64` and written straight out of one.
2. **Batch a run of adjacent fields in the derive.** Neighbouring big-endian
   `Msb0` fields that add up to 64 bits or less become a single
   read/write call plus a shift and mask per field. The 11-field CCSDS TM
   header becomes one 48-bit read.

Numbers below are per-item cost, measured on benches that push 128 frames
through a single reader or writer so nothing overlaps or gets optimised away.
Each column is the gain over the column to its left; the last is against
current `master`.

| | #666 alone | +1 over #666 | +2 over +1 | all three |
|---|---|---|---|---|
| TM header, read | 3.7x | 29x | 4.1x | **441x** |
| TM header, write | 1.0x | 2.8x | 10x | **29x** |
| 1-bit field in a `u64`, read | 31x | 41x | 1.0x | **1239x** |
| 6 plain byte fields, read | 1.0x | 1.0x | 2.7x | **2.7x** |
| 6 plain byte fields, write | 1.0x | 1.0x | 1.9x | **1.9x** |

Worth noting this PR is the only one of the three that fixes an asymptotic
problem rather than a constant factor, which is the 31x on the 1-bit case. It
also does nothing for writes, since it is a read-side fix; all the write gains
come from the follow-ups.

Is there anything you would like changed here before it can go in? And do you
have a rough idea of the timing? I would rather rebase the follow-ups onto a
merged `master` than maintain them as a stack.

wcampbell0x2a:
@Simon-Calbert-Aerospacelab I can't give you a time-table of merging. I am only the co-maintainer.

Simon-Calbert-Aerospacelab:
Follow-up to this is now open: #673. Same quadrant, one level up, it skips
the `BitSlice` entirely rather than just making the zero-extension cheap.

PR #680, opened 2026-09-03
Modernize CI and development workflow with Docker, Just, and Rust 2024

sharksforarms:
- Add a reproducible Docker Compose + Just development environment with pinned MSRV, stable, beta, embedded, and WASM toolchains (also used in CI) useful to be able to reproduce issues in CI locally
- Consolidate CI into one workflow running the full build, test, examples, no_std, and WASM pipeline across MSRV, stable, and beta in parallel (linting + coverage using stable). Also added caching.
- Migrate all crates and fixtures to Rust 2024
- Add MSRV and beta trybuild snapshots for derive-macro validation, useful to catch compiler-specific regressions earlier.

Other:
- Removed bench & miri from CI. Bench wasn't particularly useful, performance PRs include their own benches/analysis. For miri, I believe I have a plan (possibly removing bitvec in favor of our own lightweight impl)
- Updated README: See [rendered README](https://github.com/sharksforarms/deku/blob/sharksforarms/docker-just-ci/README.md)


wcampbell0x2a:
> Removed bench & miri from CI

I was actually thinking the same with my CI benchmarks being kinda useless. I went ahead and made https://github.com/wcampbell0x2a/rust-bench-compare-action#gungraun so that CI benchmarking is at-least useful. I planned on adding it here? (I used it locally on a couple of the more recent MRs)

sharksforarms:
> > Removed bench & miri from CI
> 
> I was actually thinking the same with my CI benchmarks being kinda useless. I went ahead and made https://github.com/wcampbell0x2a/rust-bench-compare-action#gungraun so that CI benchmarking is at-least useful. I planned on adding it here? (I used it locally on a couple of the more recent MRs)

sure! good idea. let's do so in a follow up

.github/workflows/main.yml
```diff
@@ -1,88 +1,169 @@
...
+        run: |
+          set -euo pipefail
+          resolve_channel() {
+            curl --fail --silent --show-error \
+              "https://static.rust-lang.org/dist/channel-rust-$1.toml" |
+              awk '/^\[pkg\.rust\]$/ { found = 1; next }
+                   found && /^version = / { gsub(/"/, "", $3); print $3; found = 0 }'
+          }
+          # extract MSRV from Cargo.toml
+          msrv=$(awk -F'"' '
```
wcampbell0x2a, line 49: I think using `cargo metadata` would be nicer here.

README.md
```diff
@@ -1,71 +1,139 @@
...
+Describe a binary layout with annotated structs and enums, then derive bit-level reading and
+writing implementations from the same definition. It is designed for network
+protocols, file formats, and embedded data.
+
+- Symmetric reading and writing from a single type definition
+- Bit-sized fields with configurable byte and bit order
+- Tagged enums, variable-length collections, and conditional fields
+- Validation, context-aware types, and custom readers and writers
+- Byte slices and `Read`/`Write` streams
+- `std`, `no_std`, and fixed-size, no-allocation workflows with `DekuSize`
```
wcampbell0x2a, line 25: Might as well add something about performant code, as it's been a goal of the project.

Dockerfile
```diff
@@ -0,0 +1,70 @@
+# syntax=docker/dockerfile:1.7
+
+FROM node:24-bookworm-slim AS node-runtime
+
+FROM rust:1.88-slim-bookworm AS deku-build-base
```
wcampbell0x2a, line 5: Can the `1.88` here be driven by the `ARG MSRV_TOOLCHAIN`?

wcampbell0x2a (changes requested):

.github/workflows/main.yml
sharksforarms: done

README.md
sharksforarms: done

Dockerfile
sharksforarms: done

README.md
```diff
@@ -1,71 +1,140 @@
...
+
+Features can be combined as needed. For example, enable `bits` for bit-level
+formats and `alloc` for heap-backed types:
+
+```toml
+[dependencies]
+deku = { version = "0.20", default-features = false, features = ["bits", "alloc"] }
+```
+
+## Development
```
sharksforarms: since this readme get's used in the crates.io.. i'm thinking of moving this out

wcampbell0x2a: I think that's fair. I just wanna remove the "good luck getting CI to pass for this MR" friction that every contributor has with this project.

PR #682, opened 2026-09-14
Add Ci grungraun; Fix 6 simple byte read regression; Add Protocol Benchmarks

wcampbell0x2a:
See commits. as they will be easier to review as commits since I moved some code around.

b923c0d bench: Use gungraun to easily compare instr counts
* Add criterion for measuring on CI (non-quiet infra). This will hopefully
  lead to good benchmarks to use in merge requests, while keeping
  the regular criterion for wall-time, which is still very usefull since
  callgrind is just estimating the cputime per instruction.
* Copy the existing tests into shared shapes usable by both.

b2e5de0 Keep whole-byte field runs on the byte path
* Fix regression fonud in be_six_bytes_aligned_x128.
* Apply batching logic only if the fields are not whole-width reading.
  For example, on a plain u8 field read with no bit-reads. This uses
  the u64 and shift/mask logic. This was greatly needed to cover
  the regression of be_six_bytes_aligned_x128. This also should work
  over the problem found in #677 regarding protocol pipeline issues.
* I benchmarked this over a multi-protocol benchmark (in which we should
  add) and it shows gains (also shows gains in the micro-benchmark
  of be_six_bytes_aligned_x128.

ccac117 Move plan_bits_run and friends into common mod

4c62abc test: Add protocol benchmarking to evade branch prediction
* Add testing to confirm the comment above read_bits_uint_msb0 where
  having code that the branch predictor can't help with quickly makes
  the code pattern that we currently have required. The code didn't really
  mention the branch predictor, but I bet that's what they were hitting.

03fe9af ci: Remove the criterion benchmark job
Keep only the callgrind benchmark, which gives stable instruction
counts. The criterion wall-clock job gave noisy results on CI runners.


wcampbell0x2a:
@Simon-Calbert-Aerospacelab in #677 you mentioned that pipeline'ing benchmarks are driving the decision you made in `src/reader.rs`, so I added some benchmarks to prove it. See https://github.com/sharksforarms/deku/commit/b2e5de0b4efab52bcb9458320a0f0240929676fc, as I make some more assumptions on getting some performance back for some common paths.

wcampbell0x2a:
The benchmark CI I added/changed is failing. But hopefully it's just benches against master which are expected to fail.

wcampbell0x2a:
I need to update for https://github.com/sharksforarms/deku/pull/684

PR #685, opened 2026-09-15
Add bound attributes to restrict derived impls

p-avital:
This is an update of #341 (as I can't edit the original, which now has conflicts according to GH; credits to @MegaBluejay for the initial implementation).

This version introduces 4 attributes: one for each of the derived traits, and an additional `bound` attribute that applies to all of them.

Allowing to place different bounds on different traits allows for asymmetric implementations, such as:

```
#[deku(write_bound = "Content<VALIDATED>: DekuWriter")]
struct Message<const VALIDATED: bool> {
  header: Header<VALIDATED>,
  content: Content,
}
```

This pattern can be useful when you want to be able to decode bit-streams that are well formed, but don't respect certain invariants; but only want to allow encoding bit-streams from values that you've already validated.

I've chosen to have the documentation for each attribute point to a single shared section to avoid adding too much noise to the documentation.


wcampbell0x2a:
@p-avital looks like some tests are failing.

wcampbell0x2a (approved):

